PR #210, opened 2016-06-08
Fix traceback when parsing ASAN logs from debug builds with logging

jschwartzentruber:
environment variables set.


choller:
Merged, Thanks!

PR #221, opened 2016-08-03
Allow EC2SpotManager to also start instance-store images.

nth10sd:
Laniakea supports instance-store images:

https://github.com/MozillaSecurity/laniakea/blob/master/laniakea.py#L67
https://github.com/MozillaSecurity/laniakea/blob/db3f397a3f78f929aa205711fb723e41fefadaa0/core/manager.py#L109

It would be nice if EC2SpotManager could start up instance-store images as well.

By default, we use EBS instances with -root-device-type=ebs, so EC2SpotManager should be able to pass in -root-device-type=instance_store if the EC2 Image Name is an instance-store image.

Note that the code in the PR does not yet work properly. I don't know how to switch the form for Root Device Type to a checkbox, but ideally we should have one probably named "Instance-store image" near these:

[View spot pool](https://github.com/MozillaSecurity/FuzzManager/blob/f58ef759db5e733aad182d27fdecc3468dbe28d3/server/ec2spotmanager/templates/config/view.html#L30)
[Edit spot pool](https://github.com/MozillaSecurity/FuzzManager/blob/f58ef759db5e733aad182d27fdecc3468dbe28d3/server/ec2spotmanager/templates/config/edit.html#L68)

Suggestions welcome!

### DIFF
--- a/server/ec2spotmanager/management/commands/start_monitoring_daemon.py
+++ b/server/ec2spotmanager/management/commands/start_monitoring_daemon.py
@@ -308,7 +308,7 @@ class Command(NoArgsCommand):
 
             try:
                 logger.info("[Pool %d] Creating %s instances..." % (pool.id, count))
-                boto_instances = cluster.create_spot(config.ec2_max_price, tags=config.ec2_tags, delete_on_termination=True, timeout=20 * 60)
+                boto_instances = cluster.create_spot(config.ec2_max_price, tags=config.ec2_tags, root_device_type=config.ec2_root_device_type, delete_on_termination=True, timeout=20 * 60)
                 canceled_requests = count - len(boto_instances)
 
                 logger.info("[Pool %d] Successfully created %s instances, %s requests timed out and were canceled" % (pool.id, len(boto_instances), canceled_requests))
--- a/server/ec2spotmanager/templates/config/edit.html
+++ b/server/ec2spotmanager/templates/config/edit.html
@@ -69,6 +69,11 @@
             <input id="id_ec2_image_name" class="form-control" name="ec2_image_name" type="text" value="{{ config.ec2_image_name|default:"" }}">
             <br/>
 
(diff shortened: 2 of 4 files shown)


choller:
If you are looking for an example on how to implement a checkbox, take a look at these two links:

https://github.com/MozillaSecurity/FuzzManager/blob/master/server/crashmanager/templates/signatures/edit.html#L27

https://github.com/MozillaSecurity/FuzzManager/blob/master/server/crashmanager/views.py#L585

Those are for handling the "frequent issue" checkbox in signatures.


Furthemore, your patch lacks the necessary changes to the database model. Keep in mind, you want to store this configuration setting somewhere, right?

1. Take a look at https://github.com/MozillaSecurity/FuzzManager/blob/master/server/ec2spotmanager/models.py and try to understand the storage model. Then add a boolean storage field to PoolConfiguration (InstancePool has one as an example). 

2. PoolConfigurations needs to be serializable and can be "flattened". You might remember from working with EC2SpotManager that configurations have parents. In the database, each of these is a PoolConfiguration, but in the end when starting an instance, you need *one* configuration. That configuration is the configuration chain flattened. What flattening does is, it converts the database object into a dict and overrides fields set in the parent with those set in the child (or for dicts, they are merged instead). For the boolean field I assume it *should* work to add it to `self.config_fields` and it will just work.

3. When that is done and working, we need a database migration. If the field has a default value (and it should, with the value False), then a migration can be created automatically with `python manage.py makemigrations`. This creates a new migration file that needs to be comitted as well. Every instance of EC2SpotManager needs to apply this configuration before continuing operations. Otherwise, you simply don't have the necessary storage field in your database.

Let me know if you have questions.

nth10sd:
I'm not sure this is needed anymore. EC2 seems to be moving towards EBS-only image types, and it seems instance store ones are on their way out. Closing for now.

PR #231, opened 2016-09-23
Fix #215 - Support Win10 crash reporter output, switch to !analyze output parsing

### LINKED ISSUE #215: Add support for !analyze output in cdb
In Windows 10, [!exploitable 1.6.0.0](https://msecdbg.codeplex.com/) is unable to get the stack - everything is listed as `??`. However, `!analyze -v` still gives us the stack, but the detection support will need to be updated.

`!analyze` is builtin whereas `!exploitable` requires installation of a separate windows debugger extension, so removing this dependency is also a good thing.

Things to do:
* Regenerate the cdb tests in FTB/Signature which adds `!analyze -v` output.
  * Initial cdb support predated the addition of `!analyze -v`, which was [added in mid-May 2016](https://github.com/MozillaSecurity/funfuzz/commit/70cca63d73fc7bb765ac021fc85499ed8d005b33).
* Update the detection support for `!analyze -v` output.

```
Edit: This section is now on hold:
( @rforbes suggested that we might be able to ask about the future of !exploitable in Windows 10 )

After FuzzManager production is updated and Win7/WinServer 2012 R2 works properly, we should:
* Remove the `!exploitable -v` command from cdbCmds.txt in the funfuzz repo.
* Stop installing !exploitable in psbootstrap.
* Regenerate the cdb tests in FTB/Signature which removes `!exploitable -v` output.
```

nth10sd:
We need to support Windows 10. Its corresponding Windows Server edition, 2016, is likely to come out end-Sep 2016 and we should be ready to support that too.

I ripped out the old tests and added new ones. The revamped tests are as follows.

Tested using the following builds from m-c rev 62f79d676e0e:

* 32-bit debug deterministic: "--enable-debug --enable-more-deterministic --32"
* 32-bit opt deterministic: "--enable-more-deterministic --32" (does not fail with assertion failure testcase)
* 64-bit debug deterministic: "--enable-debug --enable-more-deterministic"
* 64-bit opt deterministic: "--enable-more-deterministic" (does not fail with assertion failure testcase)

against an assertion failure [bug 1280710](https://bugzilla.mozilla.org/show_bug.cgi?id=1280710) and a crash [bug 1302432](https://bugzilla.mozilla.org/show_bug.cgi?id=1302432)

```
### Win10 ###

# Test 1b is for Win10 with 32-bit js debug deterministic shell hitting the assertion failure:
#     js_dbg_32_dm_windows_62f79d676e0e!js::GetBytecodeLength+47
#     01344577 cc              int     3

# Test 2b is for Win10 with 64-bit js debug deterministic shell hitting the assertion failure:
#     js_dbg_64_dm_windows_62f79d676e0e!js::GetBytecodeLength+52
#     00007ff7`1e424e62 cc              int     3

# Test 3b is for Win10 with 32-bit js debug deterministic shell crashing:
#     js_dbg_32_dm_windows_62f79d676e0e!js::gc::TenuredCell::arena+13
#     00ed6a63 8b00            mov     eax,dword ptr [eax]

# Test 4b is for Win10 with 32-bit js opt deterministic shell crashing:
#     js_32_dm_windows_62f79d676e0e!JSObject::allocKindForTenure+9
#     00404c59 8b39            mov     edi,dword ptr [ecx]

# Test 5b is for Win10 with 64-bit js debug deterministic shell crashing:
#     js_dbg_64_dm_windows_62f79d676e0e!js::gc::IsInsideNursery+1b
#     00007ff7`1dcf75db 8b11            mov     edx,dword ptr [rcx]

# Test 6b is for Win10 with 64-bit js opt deterministic shell crashing:
#     js_64_dm_windows_62f79d676e0e!JSObject::allocKindForTenure+13
#     00007ff7`4d469ff3 4c8b01          mov     r8,qword ptr [rcx]
```

For testing, I had copied the test *.txt files to the main FuzzManager directory, then ran:

```python -m unittest discover FTB/Signatures/```

and again whittled away the test failures.

I still have to test this on existing Win7 systems, will add tests later, probably as future commits. They will be tagged as 1a/2a/3a etc. Creating this PR now to first get some feedback.

Note to self: Update funfuzz and psbootstrap (e.g. removing !exploitable since there does not seem to be movement there) once this deploys to pick up the changes.

### CODE REVIEW

**Comment by nth10sd** [comment]
I've split out into a separate commit the CrashInfo.py changes from the Windows testcases, adding the Win7 testcase in the process:

```
### Win7 ###

# Test 1a is for Win7 with 32-bit js debug deterministic shell hitting the assertion failure:
#     js_dbg_32_dm_windows_62f79d676e0e!js::GetBytecodeLength
#     01814577 cc              int     3

# Test 2a is for Win7 with 64-bit js debug deterministic shell hitting the assertion failure:
#     js_dbg_64_dm_windows_62f79d676e0e!js::GetBytecodeLength
#     00000001`40144e62 cc              int     3

# Test 3a is for Win7 with 32-bit js debug deterministic shell crashing:
#     js_dbg_32_dm_windows_62f79d676e0e!js::gc::TenuredCell::arena
#     00f36a63 8b00            mov     eax,dword ptr [eax]

# Test 4a is for Win7 with 32-bit js opt deterministic shell crashing:
#     js_32_dm_windows_62f79d676e0e!JSObject::allocKindForTenure
#     00d44c59 8b39            mov     edi,dword ptr [ecx]

# Test 5a is for Win7 with 64-bit js debug deterministic shell crashing:
#     js_dbg_64_dm_windows_62f79d676e0e!js::gc::IsInsideNursery
#     00000001`3f4975db 8b11            mov     edx,dword ptr [rcx]

# Test 6a is for Win7 with 64-bit js opt deterministic shell crashing:
#     js_64_dm_windows_62f79d676e0e!JSObject::allocKindForTenure
#     00000001`3f869ff3 4c8b01          mov     r8,qword ptr [rcx]
```

File: FTB/Signatures/CrashInfo.py [hunk 1]
```diff
@@ -1148,75 +1149,88 @@ def __init__(self, stdout, stderr, configuration, crashData=None):
...
+                address = line.split()[1]
                 self.crashAddress = long(address, 16)
 
             # Crash instruction
-            if line.startswith("Faulting Instruction:"):
-                # Example:
-                #     Faulting Instruction:01206fbd int 3
-                cInstruction = line.split(":")[1].split(None, 1)[1]
+            # 64-bit binaries have a backtick in their addresses, e.g. 00007ff7`1e424e62
+            lineWithoutBacktick = line.replace("`", "")
```
choller: Is there only one backtick per line? If so, you might want to limit the replace to the first occurrence.

Also, do you need  the original line with the backtick? If not, just do line = line.replace... to keep the code simple.

FTB/Signatures/CrashInfo.py
```diff
@@ -1148,75 +1149,88 @@ def __init__(self, stdout, stderr, configuration, crashData=None):
...
+                # 64-bit example:
+                #     00007ff7`4d469ff3 4c8b01          mov     r8,qword ptr [rcx]
+                cInstruction = line.split(None, 2)[-1]
+                # There may be multiple spaces inside the faulting instruction
+                cInstruction = " ".join(cInstruction.split())
                 self.crashInstruction = cInstruction
 
             # Start of stack for crashing thread
-            if re.match(r'^\s*Hash Usage : Stack Trace:', line):
+            if line.startswith("STACK_TEXT:"):
```
choller: Is this a Win7->Win10 change or is that just a generic improvement? Does it still work with the older traces?

FTB/Signatures/tests.py
```diff
@@ -874,528 +874,782 @@ def runTest(self):
...
-        self.assertEqual(crashInfo.backtrace[43], "RunFile")
-        self.assertEqual(crashInfo.backtrace[44], "ProcessArgs")
-        self.assertEqual(crashInfo.backtrace[45], "Shell")
-        self.assertEqual(crashInfo.backtrace[46], "main")
-        self.assertEqual(crashInfo.backtrace[47], "__tmainCRTStartup")
-        self.assertEqual(crashInfo.backtrace[48], "mainCRTStartup")
-        self.assertEqual(crashInfo.backtrace[49], "BaseThreadInitThunk")
-        self.assertEqual(crashInfo.backtrace[50], "RtlInitializeExceptionChain")
-        self.assertEqual(crashInfo.backtrace[51], "RtlInitializeExceptionChain")
+        self.assertEqual(len(crashInfo.backtrace), 33)
```
choller: With such long backtraces, you don't need to check every single frame. When writing tests, you should always make sure to cover edge cases and maybe a few generic cases. E.g. you always want to check backtrace[0], you always want to check the last stackframe as well (bounds) and maybe 2-3 random frames in the middle. If there are frames with e.g. special properties or chars, that you think might be troublesome, include them. But you don't need to check them all :) You can still leave this as it is of course.

FTB/Signatures/CrashInfo.py
nth10sd: From #fuzzing discussion:
```
13:48  @gkw: decoder: it's less of Win10 vs Win7, but rather moving from !exploitable to !analysis
13:48  @gkw: the former is an extension, whereas the latter (!analysis) is builtin
13:48  <@    decoder> | then i would say we support !analysis only
13:49  <@    decoder> | if there are old entries in fuzzmanager, they will not cause problems until they are edited in some way
13:49  <@    decoder> | then they will fail to parse
13:49  <@    decoder> | but there is a cache layer that prevents them from being reparsed
13:50  @gkw: decoder: to answer your qn about why the change was needed, !exploitable was generating ?? stack entries on Win10 (the extension is not updated for Win10)
13:50  @gkw: basically none of the entries were being parsed properly as the old code relied on !exploitable output
13:50  @gkw: in Win10
13:50  <@    decoder> | well, as long as we dont want to switch back at some point to !exploitable
13:51  <@    decoder> | it would be bad to rip out a feature now if you want to go back later
13:51  <@    decoder> | but if we want to discontinue its use, that seems fine to me
13:51  @gkw: decoder: not till https://msecdbg.codeplex.com/ gets updated for Win10, which hasn't been touched in 3.5 years
13:51  @gkw: moreover we can always add !exploitable in addition to !analysis later
```
(some lines were removed to trim verbosity)

To summarise:

* The change moves to !analysis (built-in function) from !exploitable (3.5 year-old extension from https://msecdbg.codeplex.com/)
* This change will break older traces but only if they are edited in some way.
* We can always re-add support for !exploitable if needed. It hasn't been updated for 3.5 years though.

FTB/Signatures/tests.py
nth10sd: > But you don't need to check them all :) You can still leave this as it is of course.

I'll probably leave these as-is. I had bulk-copied the stack entries and "massaged" them till they fit the testing syntax, rather than manually inputting them one-by-one.

FTB/Signatures/CrashInfo.py
nth10sd: > If so, you might want to limit the replace to the first occurrence.

Good suggestion, I will do this.

> Also, do you need the original line with the backtick? If not, just do line = line.replace... to keep the code simple.

The subsequent lines operate on the assumption that the line was not changed any way, so let's keep it that way.

PR #260, opened 2016-11-11
Fix stack parsing code on Windows

### LINKED ISSUE #242: [CrashManager] Stack parsing code on Windows fails if the top line does not contain the binary name
The crash stack parsing code on Windows fails if the top line of the stack does not contain the binary name. i.e. in this case the top line contains mozglue instead.

This causes the parsing code to fail and so FuzzManager shows the Short Signature as "No crash detected", for the following example:

```
STACK_TEXT:
WARNING: Frame IP not in any known module. Following frames may be wrong.
0053e39c 746c37f5 010000f0 08900850 01000044 0x25d80b01
0053e430 014f5ac7 25d80010 00000003 0053e780 mozglue!arena_run_dalloc+0x205 [c:\users\administrator\trees\mozilla-central\memory\mozjemalloc\jemalloc.c @ 3916]
0053e508 014f78b1 08410000 0053e520 0053e6c8 js_32_windows_42c95d88aaaa!EnterIon+0xc7 [c:\users\administrator\trees\mozilla-central\js\src\jit\ion.cpp @ 2823]
0053e5a8 0136d44f 08410000 0053e668 0053e678 js_32_windows_42c95d88aaaa!js::jit::IonCannon+0x121 [c:\users\administrator\trees\mozilla-central\js\src\jit\ion.cpp @ 2921]
```

nth10sd:
This fixes #242 - the change in eip/rip extraction is needed for this. Sometimes cdb reports have weird values (e.g. unknown) when the system runs out of resources, so we have to work around this.

I've added more tests extracted from real-world results, and updated expected values in the older tests.

Running `python -m unittest discover FTB/Signatures` has been successful with this PR.


FTB/Signatures/CrashInfo.py
```diff
@@ -1168,6 +1168,15 @@ def __init__(self, stdout, stderr, configuration, crashData=None):
                     inEcxrData = False
                     continue
 
+                # Crash address
+                # Extract the eip/rip specifically for use later
+                if "eip=" in line:
+                    address = line.split("eip=")[1].split()[0]
+                    self.crashAddress = long(address, 16)
+                elif "rip" in line:
```
choller: Did you intend to use "rip=" in line here?

nth10sd:
Not sure what was needed in GitHub to approve the review since I've updated the PR and acknowledged the review.

PR #264, opened 2016-12-28
Add setup.py for client installation.

### LINKED ISSUE #263: [FTB] FTB/Collector should be easily installed by clients.
Client code which does not need to run any portion of the server should be able to easily install the client portions of FuzzManager (FTB and Collector). It should be possible to install FTB and Collector into a virtualenv -- without needing the server requirements -- either with pip by running a manual install step.

jschwartzentruber:
Fixes #263

### DIFF
--- a/requirements-client.txt
+++ b/requirements-client.txt
@@ -1,0 +1,2 @@
+numpy==1.11.2
+requests>=2.5.0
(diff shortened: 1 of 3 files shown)


choller:
Thanks for working on this!

PR #283, opened 2017-06-14
[FTB] Add support for parsing Rust backtraces.

jschwartzentruber:


choller:
We still need to make a benchmark for the regex use in  FTB/Signatures/CrashInfo.py (in fromRawCrashData) to detect which CrashInfo subclass to instantiate. If the regex is more expensive than going to a substring filter first, then we need to change that portion of the code.

Background: The fromRawCrashData function processes huge amount of data based on the submissions (e.g. if a lot of stderr is included, the server has to process every line using all of the checks in fromRawCrashData, worst case)

jschwartzentruber:
I ran this through the random logs I have locally:
* 2,037 logs
* 165,208 lines
* 200 ASan logs
* 8 Minidump logs
* 34 GDB logs
* 5 rust logs

Each loop tries to match all the logs -- the dispatch from `CrashInfo.fromRawCrashData()` only, with `CrashInfo` constructors commented out.

#### original

    10 loops, best of 3: 50.6 msec per loop

#### version checked-in using re

    10 loops, best of 3: 76.9 msec per loop

#### rust str "in" 2-lines .. "panicked at" followed by "stack backtrace:"

    10 loops, best of 3: 57.7 msec per loop

#### rust str "in" 1-line .. "thread" and "panicked at"

    10 loops, best of 3: 59.1 msec per loop

#### replacing all with a single re

    10 loops, best of 3: 163 msec per loop

I'll switch to the 2-line str test now to keep it fast and similar in style to the others. In perspective, the all regex version is probably still acceptable performance if false-positives (or negatives) become an issue.

PR #284, opened 2017-06-16
[FTB] Switch to string matching instead of regex for detecting Rust backtraces.

jschwartzentruber:
Update to #283.


jschwartzentruber:
I added some stats in #283. I guess you can't reopen PRs.

choller:
Thanks for the stats :) Seems like even though the string matching is faster it is not so much an issue as I thought.

PR #316, opened 2017-08-01
Add support for ASan negative-size-param errors

tysmith:


FTB/Signatures/CrashInfo.py
```diff
@@ -491,19 +491,22 @@ def __init__(self, stdout, stderr, configuration, crashData=None):
...
         for traceLine in asanOutput:
             if self.crashAddress == None:
                 match = re.search(asanCrashAddressPattern, traceLine)
-
                 if match != None:
-                    self.crashAddress = long(match.group(1), 16)
+                    try:
+                        self.crashAddress = long(match.group(1), 16)
+                    except TypeError:
+                        self.crashAddress = 0
```
choller: I guess this should be "None" to make sure that no crash address is recorded

FTB/Signatures/test_CrashSignature.py
```diff
@@ -523,6 +533,16 @@ def runTest(self):
...
+class SignatureStackFramesNegativeSizeParamTest(unittest.TestCase):
+    def runTest(self):
+        config = ProgramConfiguration("test", "x86-64", "linux")
+        crashInfoPos = CrashInfo.fromRawCrashData([], [], config, auxCrashData=testTraceNegativeSizeParam.splitlines())
+
+        testSig = crashInfoPos.createCrashSignature()
+
+        self.assertIn("/ERROR: AddressSanitizer", str(testSig))
+        self.assertIn("negative\\\\-size\\\\-param", str(testSig))
+        self.assertTrue(isinstance(testSig.symptoms[1], StackFramesSymptom))
```
choller: Maybe add a check that crash address is indeed not set

PR #323, opened 2017-08-09
Signatures

kkuehl:
Allow signatures.zip to be downloaded as required by funfuzz


kkuehl:
Thank you, I have reverted the previous incorrect changes and added the missing aliases.

choller:
Thanks! :)

PR #325, opened 2017-08-10
Allow downloading of crashmanager/files from funfuzz

kkuehl:


choller:
So the intended setup here is that you add a user fuzzmanager to the basic auth with the token as the password. Your patch would work here for signatures.zip but it would not work for tests (which the client can also download). For tests, the path is also used by the UI so it cannot be switched to token auth.

I am reluctant to "half-fix" this because for the full setup to work, you would still need basic auth in the client for user "fuzzmanager".

kkuehl:
Okay, using the token as the password also works. The instructions were
just unclear.

choller:
Yea, sorry for that :( The documentation is really not up-to-date in some places and also misses some edge cases. If you want to add something about this to the documentation and update this PR, then I would be very happy to merge it. You should try rebasing first though to make sure you don't have more commits in the PR (or open a new one from a separate branch).

Thanks for trying our FuzzManager and helping with the docs! :)

PR #328, opened 2017-08-11
Update README.md and add new fuzzmanager_setup_helper.py script.

kkuehl:
See the example output in README.md for fuzzmanager_setup_helper.py usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,9 @@ Afterwards, you should run the following commands
 
 ```
 $ cd server
-$ python manage.py syncdb
+$ python manage.py makemigrations
+$ python manage.py migrate
+
 ```
 During syncdb, make sure that you create a user as we need it later on.
 


kkuehl:
I cancelled my last PR and based upon your feedback, reverted those changes, updated the README.md and created a new helper script to assist others.

README.md
```diff
@@ -64,7 +64,10 @@ Afterwards, you should run the following commands
 $ cd server
 $ python manage.py syncdb
 ```
-During syncdb, make sure that you create a user as we need it later on.
+Manually create the fuzzmanager user
```
choller: I don't remember creating this user manually but maybe I logged in with the user to create it. Anyway, the correct way to fix this is to use get_or_create for the user in get_auth_token.py. Something like

`(user, _) = User.objects.get_or_create(username=label)`

in server/crashmanager/management/commands/get_auth_token.py should fix the issue. I can fix this, or you can update your PR, as you want :)

README.md
```diff
@@ -90,6 +93,19 @@ Use the following command to get an authentication token for a Django user:
...
 `python manage.py get_auth_token username`
 
+Obtain the fuzzmanager auth_token to be used in .fuzzmanagerconf as well as the basic authentication password
+
+    python manage.py get_auth_token fuzzmanager
+    4a253efa90f514bd89ae9a86d1dc264aa3133945
+   
+Set fuzzmanager's http basic authentication password to the auth_token 
+
+`httpasswd -cb .htpasswd fuzzmanager 4a253efa90f514bd89ae9a86d1dc264aa3133945`
```
choller: I think there's a typo in "httpasswd"

examples/apache2/default.vhost
```diff
@@ -21,7 +21,7 @@ WSGIApplicationGroup %{GLOBAL}
 		# Your LDAP configuration here, including Require directives
 
 		# This user is used by clients to download test cases and signatures
-		Require user fuzzmanager
+		Require valid-user
```
choller: If you change this, can you also update the comment above to something like "Users need to be able to download ..."?

server/fuzzmanager_setup_helper.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python
```
choller: Thanks for contributing this helper script :) Could you maybe make a new directory server/contrib/ and move the script there? Then move the script output from README.md to a new README.md in server/contrib/ so we don't clutter the main README too much. I will add a reference to your script in main README.md myself and also edit the contrib README later on so we can add more user-contributed setup helpers. Thanks again!

kkuehl:
I have addressed your requested changes. Thanks!

PR #340, opened 2017-08-30
update ec2reporter.py with disable flag

rforbes:


EC2Reporter/EC2Reporter.py
```diff
@@ -152,6 +152,21 @@ def cycle(self, poolid):
...
         if response.status_code != requests.codes["ok"]:
             raise self.__serverError(response)
+            
+    @remote_checks
+    def disable(self, poolid):
+        '''
+        Disable the pool with the given id.
+        
+        @type poolid: int
+        @param poolid: ID of the pool to cycle
```
choller: Param docstring looks outdated

EC2Reporter/EC2Reporter.py
```diff
@@ -172,6 +187,7 @@ def main(argv=None):
     actions.add_argument("--report", dest="report", type=str, help="Submit the given textual report", metavar="TEXT")
     actions.add_argument("--report-from-file", dest="report_file", type=str, help="Submit the given file as textual report", metavar="FILE")
     actions.add_argument("--cycle", dest="cycle", type=str, help="Cycle the pool with the given ID", metavar="ID")
+    actions.add_argument("--disable", dest="disable", type=str, help="Disable the pool with the given ID", metavar="ID")
```
choller: If you add --disable, please also add --enable. We need both anyway for the coverage stuff if we don't want to manually kick off coverage generation.

EC2Reporter/EC2Reporter.py
choller: Param docstring looks outdated

choller:
Requested changes are addressed in PR #341

Issue #235, opened 2016-09-28
[CrashManager] toolfilter does not work with the /?all=1 page


nth10sd:
@choller mentioned over IRC that toolfilter is right now not intended to work on the /all/ page but looking at the code, it should work with /?all=1 - but doesn't yet.

choller:
We need to figure out first what is the intended behavior here, then the fix should be simple.

nth10sd:
The crashes page has indeed changed but I think I was thinking about the `signatures/all/?all=1` vs `signatures/all` page instead.

nth10sd:
@jschwartzentruber and I had a discussion and my issue can be summarised as wanting to see:

`All signatures with tool filter on, independent of presence of bug number`

Right now, `signatures/?all=1` does not seem to work properly. There are entries shown that when one clicks into the individual reports within the signatures/buckets, it shows 0 entries (due to the toolfilter). These entries should not be shown in `signatures/?all=1` in the first place.

jschwartzentruber:
I think the intention is this:

| url | use toolfilter | show bucketed |
|---|---|---|
| `crashes/` | yes | no |
| `crashes/?all=1` | yes | yes |
| `crashes/all/` | no | no |
| `crashes/all/?all=1` | no | yes |

| url | use toolfilter | show logged |
|---|---|---|
| `signatures/` | yes | no |
| `signatures/?all=1` | yes | yes |
| `signatures/all/` | no | no |
| `signatures/all/?all=1` | no | yes |

The parameters here are very confusing. I'd propose the following:

| url | new |
|---|---|
| `crashes/` | `crashes/` |
| `crashes/?all=1` | `crashes/?show-bucketed=1` |
| `crashes/all/` | `crashes/?ignore-toolfilter=1` |
| `crashes/all/?all=1` | `crashes/?ignore-toolfilter=1&show-bucketed=1` |
| `signatures/` | `signatures/` |
| `signatures/?all=1` | `signatures/?show-logged=1` |
| `signatures/all/` | `signatures/?ignore-toolfilter=1` |
| `signatures/all/?all=1` | `signatures/?ignore-toolfilter=1&show-logged=1` |

This is apart from the fact that some cases may not work, but at least then it's clear what the intentions are and we can write tests for them.

jschwartzentruber:
These views have all been rewritten since this time.

PR #351, opened 2017-09-21
Fix the crashes/all/ view so filtering works as expected.

jschwartzentruber:


jschwartzentruber:
I think this fixes #235

PR #354, opened 2017-09-29
Implement continuous testing of server and client code.

jschwartzentruber:
Current results are here:
[travis](https://codecov.io/gh/jschwartzentruber/FuzzManager/branch/srvtests)
[codecov](https://travis-ci.org/jschwartzentruber/FuzzManager)


FTB/Signatures/test_CrashInfo.py
```diff
@@ -20,6 +21,8 @@
     NoCrashInfo, MinidumpCrashInfo, AppleCrashInfo, CDBCrashInfo, RustCrashInfo
 from FTB.Signatures.CrashSignature import CrashSignature
 
+CWD = os.path.dirname(__file__)
```
choller: Afaik, this needs to be `os.path.dirname(os.path.realpath(__file__))` instead to properly cope with relative paths and/or symlinks.

server/covmanager/SourceCodeProvider/test_sourcecodeprovider.py
```diff
@@ -88,6 +88,7 @@ def runTest(self):
         parents = provider.getParents("c3abaa766d52")
         self.assertEqual(len(parents), 0)
 
+@unittest.skipIf(not os.path.isdir("/home/decoder/Mozilla/repos/mozilla-central-fm"), reason="not decoder")
```
choller: ( ╯°□°)╯ ┻━━┻ I didn't even mean to check this in. But it's ok to disable it for now until we can properly write a test. And this is a really creative way to do so ;D

server/covmanager/views.py
```diff
@@ -349,8 +349,7 @@ class CollectionViewSet(mixins.CreateModelMixin,
         CollectionFilterBackend
     ]
 
-class RepositoryViewSet(mixins.CreateModelMixin,
-                        mixins.ListModelMixin,
+class RepositoryViewSet(mixins.ListModelMixin,
```
choller: Nice, I guess this allowed creating repositories remotely before your change?

FTB/Signatures/test_CrashInfo.py
jschwartzentruber: Possibly, yeah. I'll merge and then push this fix to master.

server/covmanager/views.py
jschwartzentruber: Yeah.. It isn't a terrible thing but I didn't think it was intended.

PR #364, opened 2017-10-11
[FTB] Remove the major source of recursion in stack symptom matching.

jschwartzentruber:
Instead of recursing into every frame on wildcards, consume the frame and continue in an outer loop. This still recurses once per wildcard to check if the rest of the stack matches, but will avoid "too much recursion" exceptions when processing very large stacks. Note that all other branches in the `_match` function already return, so continuing from this case has the same effect as recursing.

Sorry about the whitespace.


jschwartzentruber:
Yeah, just makes diffs harder to review when whitespace changes. I should probably just run sed on the whole codebase :)

This is passing the test suite and codecov says the function is 100% covered.

PR #368, opened 2017-10-11
python manage.py syncdb is no longer valid after django framework upgrade.

kkuehl:
syncdb was removed. When pip install -r server/requirements.txt is performed, the current README.md instructions are invalid due to Django
framework upgrade.

server/files directory didn't exist by default which would caus crontab
jobs as written in the README.md to fail, files/README is put in place
so the directory can be committed.
added server/contrib which contains helper scripts to ease in the setup
process.
replaced python ./manage syncdb with python ./manage.py migrage

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,9 +65,33 @@ Afterwards, you should run the following commands
 
 ```
 $ cd server
-$ python manage.py syncdb
+$ python manage.py migrate
 ```
-During syncdb, make sure that you create a user as we need it later on.
+
+Create the fuzzmanager user.
+```
+$ python ./manage.py createsuperuser
+Username (leave blank to use 'kkuehl'): fuzzmanager
+Email address: fuzzmanager@internal.com
+Password:
+Password (again):
+Superuser created successfully.
+```
+Get fuzzmanager authorization token
+```
+$ python manage.py get_auth_token fuzzmanager
+4a253efa90f514bd89ae9a86d1dc264aa3133945
+```
+Since the fuzzmanager account is used as a service account, we need to set the http basic authentication password to the auth token.
+```
+htpasswd -cb .htpasswd fuzzmanager 4a253efa90f514bd89ae9a86d1dc264aa3133945`
+```
+This .htpasswd file can be stored anywhere on your hard drive.
+Your Apache AuthUserFile line should be updated to reflect your path.
+See examples/apache2/default.vhost for an example
+
+
+
 
 ### Local testing
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ $ python manage.py migrate
 Create the fuzzmanager user.
 ```
 $ python ./manage.py createsuperuser
-Username (leave blank to use 'kkuehl'): fuzzmanager
+Username (leave blank to use 'user'): fuzzmanager
 Email address: fuzzmanager@internal.com
 Password:
 Password (again):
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -65,9 +65,33 @@ Afterwards, you should run the following commands
...
 $ cd server
-$ python manage.py syncdb
+$ python manage.py migrate
 ```
-During syncdb, make sure that you create a user as we need it later on.
+
+Create the fuzzmanager user.
+```
+$ python ./manage.py createsuperuser
+Username (leave blank to use 'kkuehl'): fuzzmanager
```
jschwartzentruber: Can you change this to 'user' ?

server/contrib/create_user.py
```diff
@@ -0,0 +1,45 @@
...
+
+import os
+import sys
+import argparse
+import random
+import string
+import django
+
+
+def main(arguments):
```
jschwartzentruber: `arguments` is unused here. Either pass it to `parse_args()` or remove it.

server/contrib/create_user.py
```diff
@@ -0,0 +1,45 @@
...
+    args = parser.parse_args()
+
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "server.settings")
+
+    from django.contrib.auth.models import User
+    django.setup()
+
+    username = args.username
+    email = args.email
+    password = ''.join(random.sample(string.letters, 20)) if args.password is None else args.password
```
jschwartzentruber: Nice, this is clean :)

server/contrib/create_user.py
```diff
@@ -0,0 +1,45 @@
...
+        user_obj = User.objects.get(username=args.username)
+        user_obj.set_password(password)
+        user_obj.save()
+    except User.DoesNotExist:
+        if superuser:
+            User.objects.create_superuser(username, email, password)
+        else:
+            User.objects.create_user(username, email, password)
+
+    print password
```
jschwartzentruber: We're trying to use `from __future__ import print_function` but this can easily be changed later in a big batch change if you don't want to update it.

server/contrib/fuzzmanager_setup_helper.py
```diff
@@ -0,0 +1,106 @@
...
+import random
+import string
+import django
+import subprocess
+
+
+def create_fuzzmanager():
+    # This is a throw away password, that is soon reset to the auth_token for fuzzmanager
+    password = ''.join(random.sample(string.letters, 20))
+    print '[+] Creating fuzzmanager user account.'
```
jschwartzentruber: Ditto re: `from __future__ import print_function` but again I'd merge it as is.

server/contrib/fuzzmanager_setup_helper.py
```diff
@@ -0,0 +1,106 @@
...
+        print '[+] The fuzzmanager account password has been set to the auth_token {}'.format(auth_token)
+        output = subprocess.check_output(['htpasswd', '-cb', '.htpasswd', 'fuzzmanager', auth_token])
+        print output
+        print '[+] .htpasswd created for apache basic authentication with the fuzzmanager user password set to {}'.format(auth_token)
+        print 'Update the AuthUserFile path/to/.htpasswd line as shown in examples/apache2/default.vhost to point to this file'   
+    except User.DoesNotExist:
+            print 'Something went wrong creating the fuzzmanager account'
+
+
+def main(arguments):
```
jschwartzentruber: Ditto re: `arguments`. They aren't used so why pass them?

kkuehl:
I have made the changes based on your feedback. Thanks!

PR #369, opened 2017-10-12
[FTB] Make signature creation use less specific addresses

tysmith:
This will help with local bucketing to reduce load on the server.

### DIFF
--- a/FTB/Signatures/CrashInfo.py
+++ b/FTB/Signatures/CrashInfo.py
@@ -391,11 +391,8 @@ class CrashInfo():
                 # Try to match crash addresses that are small but non-zero
                 # with a generic range that is likely associated with null-deref.
                 crashAddress = "< 0x100"
-            elif self.crashAddress > 0x7ff000000000:
-                # Try to match large addresses
-                crashAddress = "> 0x7ff000000000"
             else:
-                crashAddress = "> 0x100"
+                crashAddress = "> 0xFF"
 
             crashAddressSymptomObj = { "type" : "crashAddress", "address" : crashAddress }
             symptomArr.append(crashAddressSymptomObj)
--- a/FTB/Signatures/CrashInfo.py
+++ b/FTB/Signatures/CrashInfo.py
@@ -385,14 +385,17 @@ class CrashInfo():
         includeCrashInstruction = (stackIsInsufficient and self.crashInstruction != None) or forceCrashInstruction
 
         if includeCrashAddress:
-            if self.crashAddress == None:
+            if self.crashAddress is None:
                 crashAddress = ""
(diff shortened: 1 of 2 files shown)


jschwartzentruber:
@choller I encouraged this since we frequently see variable high addresses and in firefox matching on an exact address is rarely helpful. Maybe this is not the case with jsshell. The failing test uses a repeating bit pattern as an exact address match, which we could special case like this:

    address > 0x100 and re.match(r"^0x(%02x)+L?$" % (address & 0xFF), hex(address))

I also don't think matching on exactly 0x0 on firefox is useful, since we're usually access objects and this just means we accessed the first member (refcount?) but we see duplicates with low addresses in these cases which are probably the same bug. Again this might be firefox specific.

In any case, I think the 0x100 limit for low addresses is very conservative. Do you see a problem with moving this to 0x300 which would catch more accesses on null pointers to bigger object types?

choller:
First of all, thanks for working on this! This was on my todo list for a long time, I think we even have a ticket for it somewhere.

As for the details: Matching on 0x0 actually has a defined meaning on Firefox (because all aborts are 0x0 typically), but we can easily change that to be always < 0xlimit. As for the limit itself, I would rather stay at 0x100 (and yes, this is conservative). The reason is that we have seen security bugs in the past with e.g. 0x224 where the address was really controllable, but I have rarely seen anything above 0x100 that was a regular null-deref (and for the case where this happens, you can still just change the signature after verifying that it is such a case).

In general, we often rely on the address as a distinguishing element for security vs. regular bucket (e.g. you have a stack that is quite generic, with a crash at 0x4, then you rely on the combination of both to be a regular bug but you explicitly want nothing to go in there that deviates strongly from that address). Hence I would like to keep the default proposed address there at a conservative limit and rather increase it manually only when needed.

I'll add the review in a few.

tysmith:
I like the "near stack end" idea. I have removed that part for now and fixed the tests.

choller:
Thanks!

If you want to implement the near-stack-limit directive for crashAddress (in another PR), here's some pointers on how to do that:

Probably needs a special case here to check for "near-stack-limit" instead of just assuming it is a NumberMatch (and corresponding changes in the matches method of that class):

https://github.com/MozillaSecurity/FuzzManager/blob/master/FTB/Signatures/Symptom.py#L190

The logic that decides if a given address is near the stack limit should probably be implemented in a new method added to this file (because it already has architecture-specific things):

https://github.com/MozillaSecurity/FuzzManager/blob/master/FTB/Signatures/RegisterHelper.py

Adding this automatically to the proposed signature when generating goes to the same file you already modified for this PR.

The only thing we need to decide is how close to the stack limit we want to consider "near", to avoid false positives.

Issue #377, opened 2017-10-18
DATA_UPLOAD_MAX_MEMORY_SIZE default too small


kkuehl:
[Wed Oct 18 20:03:24.388579 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195] mod_wsgi (pid=12): Exception occurred processing WSGI script '/web/server/server/wsgi.py'.
[Wed Oct 18 20:03:24.388617 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195] Traceback (most recent call last):
[Wed Oct 18 20:03:24.388633 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195]   File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/wsgi.py", line 157, in __call__
[Wed Oct 18 20:03:24.388675 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195]     response = self.get_response(request)
[Wed Oct 18 20:03:24.388684 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195]   File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 124, in get_response
[Wed Oct 18 20:03:24.388699 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195]     response = self._middleware_chain(request)
[Wed Oct 18 20:03:24.388708 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195]   File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/exception.py", line 43, in inner
[Wed Oct 18 20:03:24.388722 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195]     response = response_for_exception(request, exc)
[Wed Oct 18 20:03:24.388731 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195]   File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/exception.py", line 83, in response_for_exception
[Wed Oct 18 20:03:24.388744 2017] [wsgi:error] [pid 12:tid 140541535844096] [remote 172.19.0.1:5195]     response = debug.technical_500_response(request, *sys.exc_info(), status_code=400)
[...]

PR #379, opened 2017-10-18
Fix #376 and #377

### LINKED ISSUE #376: django settings ALLOWED_HOSTS = []
/FuzzManager/server/server/settings.py
should have 
ALLOWED_HOSTS = ['*']
instead of 
ALLOWED_HOSTS = [ ]

or you get the following error:
w[Wed Oct 18 20:05:00.978014 2017] [wsgi:error] [pid 12:tid 140541544236800] Invalid HTTP_HOST header: '10.0.0.1:8000'. You may need to add u'10.0.0.1' to ALLOWED_HOSTS.

Or at least add documentation to tell the user to change this according to:
https://docs.djangoproject.com/en/1.11/ref/settings/#allowed-hosts

kkuehl:


jschwartzentruber:
These are both configuration specific, so I'd rather update `README.md` than make the defaults very permissive. We could also put comments in `settings.py` with your suggestions.

I didn't know about `DATA_UPLOAD_MAX_MEMORY_SIZE` and I'm surprised we haven't hit that yet since the upgrade. `ALLOWED_HOSTS` at least has a very clear error message if misconfigured.

kkuehl:
Reverted settings.py changes and added section to README.md

README.md
```diff
@@ -104,6 +104,20 @@ This .htpasswd file can be stored anywhere on your hard drive.
 Your Apache AuthUserFile line should be updated to reflect your path.
 See examples/apache2/default.vhost for an example
 
+### Important changes in settings.py
+It is important that you edit FuzzManager/server/settings.py and adjust the following variables according to your needs.
+
+ALLOWED_HOSTS = []
```
jschwartzentruber: This should be indented 4 spaces to be formatted properly.

README.md
```diff
@@ -104,6 +104,20 @@ This .htpasswd file can be stored anywhere on your hard drive.
...
+
+See [ALLOWED_HOSTS documentation](https://docs.djangoproject.com/en/1.11/ref/settings/#allowed-hosts)
+
+
+You may also want to increase the maximum size in bytes allowed in a request body. The default of 2.5MB may not be enough
+in some cases by adding the following variable.
+
+DATA_UPLOAD_MAX_MEMORY_SIZE = <YOUR VALUE HERE>
```
jschwartzentruber: Ditto here. `<YOUR VALUE HERE>` doesn't actually show in the markdown preview.

kkuehl:
Fixed.

Issue #383, opened 2017-10-30
MultipleObjectsReturned: get() returned more than one Product -- it returned 2!


kkuehl:
[Mon Oct 30 05:10:56.021729 2017] [wsgi:error] [pid 12:tid 140702473750272] Internal Server Error: /crashmanager/rest/crashes/
[Mon Oct 30 05:10:56.021740 2017] [wsgi:error] [pid 12:tid 140702473750272] Traceback (most recent call last):
[Mon Oct 30 05:10:56.021747 2017] [wsgi:error] [pid 12:tid 140702473750272]   File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/exception.py", line 41, in inner
[Mon Oct 30 05:10:56.021753 2017] [wsgi:error] [pid 12:tid 140702473750272]     response = get_response(request)
[Mon Oct 30 05:10:56.021768 2017] [wsgi:error] [pid 12:tid 140702473750272]   File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
[Mon Oct 30 05:10:56.021775 2017] [wsgi:error] [pid 12:tid 140702473750272]     response = self._get_response(request)
[Mon Oct 30 05:10:56.021783 2017] [wsgi:error] [pid 12:tid 140702473750272]   File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 187, in _get_response
[Mon Oct 30 05:10:56.021790 2017] [wsgi:error] [pid 12:tid 140702473750272]     response = self.process_exception_by_middleware(e, request)
[Mon Oct 30 05:10:56.021797 2017] [wsgi:error] [pid 12:tid 140702473750272]   File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 185, in _get_response
[Mon Oct 30 05:10:56.021805 2017] [wsgi:error] [pid 12:tid 140702473750272]     response = wrapped_callback(request, *callback_args, **callback_kwargs)
[Mon Oct 30 05:10:56.021812 2017] [wsgi:error] [pid 12:tid 140702473750272]   File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
[Mon Oct 30 05:10:56.021818 2017] [wsgi:error] [pid 12:tid 140702473750272]     return view_func(*args, **kwargs)
[...]

PR #386, opened 2017-11-03
Multiple objects returned

kkuehl:


jschwartzentruber:
See #383. This is a config error, and we should make the error message more explicit instead of allowing it to pass.

PR #435, opened 2018-02-15
Initial port to Python 3.

jschwartzentruber:
All tests are passing in 2.7 & 3.4-3.6. I've been running this on my test server on Python 3.5. It may not be complete but it doesn't seem to break anything in regular use under either 2.7 or 3.5 at least.

I added the [six](https://pypi.python.org/pypi/six/1.11.0) library as a dependency to aid in 2to3 issues like unicode, metaclasses, and moved imports.


server/crashmanager/serializers.py
```diff
@@ -91,14 +92,12 @@ def create(self, attrs):
...
-                testcase = base64.b64decode(testcase)
-
             h = hashlib.new('sha1')
             if testcase_isbinary:
-                h.update(str(testcase))
+                testcase = base64.b64decode(testcase)
+                h.update(testcase)
             else:
-                h.update(repr(testcase))
+                h.update(repr(testcase).encode("utf-8"))
```
choller: Can you please check that this doesn't break anything? We had a history with this line breaking due to encoding problems of plaintext tests.

PR #449, opened 2018-03-09
Add support for line-endings in EC2 user_data.

jschwartzentruber:
EC2 user_data can be either Unix or DOS line-endings. We should support both, since Windows can also use user_data scripts.


jschwartzentruber:
We discussed always converting to Unix line-endings in laniakea, but I think it should remain agnostic. We may use this for Windows user_data in the future which should keep DOS line endings.

PR #489, opened 2018-06-22
Separate out EC2 functionality

rforbes:


server/ec2spotmanager/tasks.py
```diff
@@ -6,19 +6,18 @@
...
 
-
 logger = logging.getLogger("ec2spotmanager")
 
-
 SPOTMGR_TAG = "SpotManager"
 
-
 @app.task
 def check_instance_pool(pool_id):
```
jschwartzentruber: There needs to be some abstraction here. This task has parts which will be generic for an `InstancePool` of any provider, and parts which are EC2 specific.

server/ec2spotmanager/urls.py
```diff
@@ -18,12 +18,12 @@
...
-    url(r'^configurations/(?P<configid>\d+)/$', views.viewConfig, name='configview'),
-    url(r'^configurations/(?P<configid>\d+)/edit/$', views.editConfig, name='configedit'),
-    url(r'^configurations/(?P<configid>\d+)/delete/$', views.deleteConfig, name='configdel'),
-    url(r'^configurations/(?P<configid>\d+)/cycle/$', views.forceCyclePoolsByConfig, name='configcycle'),
+    url(r'^Ec2configurations/$', views.viewEc2Configs, name='configs'),
+    url(r'^Ec2configurations/create/$', views.createEc2Config, name='configcreate'),
+    url(r'^Ec2configurations/(?P<configid>\d+)/$', views.viewEc2Config, name='configview'),
+    url(r'^Ec2configurations/(?P<configid>\d+)/edit/$', views.editEc2Config, name='configedit'),
+    url(r'^Ec2configurations/(?P<configid>\d+)/delete/$', views.deleteEc2Config, name='configdel'),
+    url(r'^Ec2configurations/(?P<configid>\d+)/cycle/$', views.forceCyclePoolsByConfig, name='configcycle'),
```
jschwartzentruber: The URLs should not be changing wholesale. As much as possible, we should not duplicate everything for each provider.

server/ec2spotmanager/views.py
```diff
@@ -242,20 +242,21 @@ def recurse_get_dependent_configurations(config):
 def createPool(request):
     if request.method == 'POST':
         pool = InstancePool()
-        config = get_object_or_404(PoolConfiguration, pk=int(request.POST['config']))
+        config = get_object_or_404(Ec2PoolConfiguration, pk=int(request.POST['config']))
```
jschwartzentruber: The specific type of `PoolConfiguration` is irrelevant here.

server/ec2spotmanager/views.py
```diff
@@ -242,20 +242,21 @@ def recurse_get_dependent_configurations(config):
...
         pool.config = config
         pool.save()
         return redirect('ec2spotmanager:poolview', poolid=pool.pk)
     elif request.method == 'GET':
-        configurations = PoolConfiguration.objects.all()
+        configurations = Ec2PoolConfiguration.objects.all()
```
jschwartzentruber: This is incorrect: creating an `InstancePool` should work for all providers.

server/ec2spotmanager/views.py
```diff
@@ -242,20 +242,21 @@ def recurse_get_dependent_configurations(config):
...
         return render(request, 'pools/create.html', {'configurations': configurations})
     else:
         raise SuspiciousOperation
 
 
 @deny_restricted_users
-def viewConfigs(request):
+def viewEc2Configs(request):
```
jschwartzentruber: The list of `PoolConfigurations` should include all providers.

server/ec2spotmanager/views.py
```diff
@@ -274,20 +275,20 @@ def add_children(node):
 
 
 @deny_restricted_users
-def viewConfig(request, configid):
-    config = get_object_or_404(PoolConfiguration, pk=configid)
+def viewEc2Config(request, configid):
+    config = get_object_or_404(Ec2PoolConfiguration, pk=configid)
```
jschwartzentruber: The abstraction is in the wrong place here. The URL and handler should be common, the template should be called `view_ec2.html`, and we should select which template to render based on the specific type of the `PoolConfiguration`.

server/ec2spotmanager/views.py
```diff
@@ -274,20 +275,20 @@ def add_children(node):
...
 
     data = {'config': config}
 
     return render(request, 'config/view.html', data)
 
 
-def __handleConfigPOST(request, config):
+def __handleEc2ConfigPOST(request, config):
```
jschwartzentruber: Capitalize EC2.

server/ec2spotmanager/views.py
```diff
@@ -378,20 +379,20 @@ def __handleConfigPOST(request, config):
 
 
 @deny_restricted_users
-def createConfig(request):
+def createEc2Config(request):
```
jschwartzentruber: In this case we probably do need separate URLs (eg. `configurations/create/ec2/`), but the handler can be the same.

server/ec2spotmanager/views.py
```diff
@@ -378,20 +379,20 @@ def __handleConfigPOST(request, config):
...
     if request.method == 'POST':
-        config = PoolConfiguration()
-        return __handleConfigPOST(request, config)
+        config = Ec2PoolConfiguration()
+        return __handleEc2ConfigPOST(request, config)
```
jschwartzentruber: This should depend on which type of configuration the user wanted to create.

server/ec2spotmanager/views.py
```diff
@@ -378,20 +379,20 @@ def __handleConfigPOST(request, config):
...
     elif request.method == 'GET':
-        configurations = PoolConfiguration.objects.all()
+        configurations = Ec2PoolConfiguration.objects.all()
```
jschwartzentruber: This is used to render the list of parent configurations. We should allow a parent configuration to be same or super type.

server/ec2spotmanager/views.py
```diff
@@ -378,20 +379,20 @@ def __handleConfigPOST(request, config):
...
 
         if "clone" in request.GET:
-            config = get_object_or_404(PoolConfiguration, pk=int(request.GET["clone"]))
+            config = get_object_or_404(Ec2PoolConfiguration, pk=int(request.GET["clone"]))
```
jschwartzentruber: The type of the `config` object should depend on what the user is trying to create.

server/ec2spotmanager/views.py
```diff
@@ -378,20 +379,20 @@ def __handleConfigPOST(request, config):
...
             config.name = "%s (Cloned)" % config.name
             config.pk = None
             clone = True
         else:
-            config = PoolConfiguration()
+            config = Ec2PoolConfiguration()
```
jschwartzentruber: The type of this object should depend on what the user is trying to create.

server/ec2spotmanager/views.py
```diff
@@ -407,14 +408,14 @@ def createConfig(request):
 
 
 @deny_restricted_users
-def editConfig(request, configid):
-    config = get_object_or_404(PoolConfiguration, pk=configid)
+def editEc2Config(request, configid):
```
jschwartzentruber: This URL and handler should be generic, but use the type of the configuration object to decide which configuration class to use.

server/ec2spotmanager/views.py
```diff
@@ -471,8 +472,8 @@ def deletePoolMsg(request, msgid):
 
 
 @deny_restricted_users
-def deleteConfig(request, configid):
-    config = get_object_or_404(PoolConfiguration, pk=configid)
+def deleteEc2Config(request, configid):
```
jschwartzentruber: This URL and handler should be generic. You shouldn't have to change anything.

rforbes:
choller=r?

server/ec2spotmanager/cron.py
```diff
@@ -34,10 +34,10 @@ def update_stats():
             current_delta_entry = PoolUptimeDetailedEntry()
             current_delta_entry.pool = pool
 
-        current_delta_entry.target = pool.config.flatten().size
+        current_delta_entry.target = pool.config.provider.flatten().size
 
-        actual = Instance.objects.filter(pool=pool).filter(Q(status_code=INSTANCE_STATE['pending']) |
-                                                           Q(status_code=INSTANCE_STATE['running'])).count()
+        actual = EC2Instance.objects.filter(pool=pool).filter(Q(status_code=EC2_INSTANCE_STATE['pending']) |
+                                                           Q(status_code=EC2_INSTANCE_STATE['running'])).count()
```
choller: This should later probably be abstracted in such a way that the specific instance model decides when machines are live/pending etc., and this code calls the abstraction for that. The notion of pending, running, etc is not EC2 specific.

server/ec2spotmanager/tasks.py
```diff
@@ -38,87 +38,90 @@ def check_instance_pool(pool_id):
...
-                running_instances.append(instance)
-            elif instance.status_code in [INSTANCE_STATE['shutting-down'], INSTANCE_STATE['terminated']]:
-                # The instance is no longer running, delete it from our database
-                logger.info("[Pool %d] Deleting terminated instance with EC2 ID %s from our database.",
-                            instance_pool.id, instance.ec2_instance_id)
-                instance.delete()
+        config = instance_pool.config.provider.flatten()
+
+        # the vast majority of this is ec2 specific. We will need to add task support for other providers
+        if config.ec2poolconfiguration:
```
choller: Remove this diff and all of the diffs in this file that start with "if config.ec2poolconfiguration:" because this is not the right way to add support for other providers. Instead, for now add a hard check that the config is actually EC2 and raise a RuntimeException if it isn't.

Later, you will have to analyze what particular tasks are being done here by this code, abstract them, and move their specialized implementations into the model classes.

server/ec2spotmanager/tasks.py
```diff
@@ -234,389 +230,438 @@ def _start_pool_instances(pool, config, count=1):
...
-            logger.error("[Pool %d] Failed to compile userdata.", pool.id)
 
+            logger.error("[Pool %d] Failed to compile userdata.", pool.id)
             entry = PoolStatusEntry()
-            entry.type = POOL_STATUS_ENTRY_TYPE['config-error']
+            entry.type = EC2_POOL_STATUS_ENTRY_TYPE['config-error']
             entry.pool = pool
             entry.isCritical = True
-            entry.msg = "Configuration error: Failed to compile userdata"
+            entry.msg = "Configuration error: %s" % traceback.format_exc()
```
choller: Do not expose a traceback here. They disclose server paths and are unsuitable for production error messages. Instead, you can output the backtrace if the Django instance is running in DEBUG mode, or output it to one of the configured log files.

server/ec2spotmanager/models.py
```diff
@@ -218,7 +254,7 @@ def getMissingParameters(self):
         return missing_fields
 
 
-@receiver(models.signals.post_delete, sender=PoolConfiguration)
+@receiver(models.signals.post_delete, sender=PoolConfiguration.provider)
```
choller: I doubt this works and you reported that file deletion didn't work in your environment. sender here must be a class, not a class field.

server/ec2spotmanager/models.py
```diff
@@ -97,29 +137,32 @@ def __init__(self, *args, **kwargs):
         # automatically here. You need to explicitly call the
         # deserializeFields method if you need this data.
 
-        super(PoolConfiguration, self).__init__(*args, **kwargs)
 
     def flatten(self):
```
choller: Why is there a flatten in the parent, and another flatten here? Shouldn't flatten be provider-independent except for the list of non-scalar fields (dicts, lists)?

jschwartzentruber:
Fixed in #510.

PR #495, opened 2018-07-23
[CovManager] Add initial support for aggregating coverage collections

choller:


server/covmanager/tasks.py
```diff
@@ -22,3 +26,39 @@ def check_revision_update(pk):
...
+    for collection in collections[1:]:
+        # Load coverage, perform the merge, then release reference to the JSON blob again
+        collection.loadCoverage()
+        CoverageHelper.merge_coverage_data(newCoverage, collection.content)
+        collection.content = None
+
+    # Save the new coverage blob to disk and database
+    newCoverage = json.dumps(newCoverage, separators=(',', ':'))
+    h = hashlib.new('sha1')
+    h.update(repr(newCoverage).encode('utf-8'))
```
jschwartzentruber: Using `repr()` on this seems unnecessary. We just got it from `json.dumps()`, so we know it's a string.

server/covmanager/templates/collections/index.html
```diff
@@ -229,11 +231,67 @@
...
+            sweetAlert('Oops', E_SERVER_ERROR, 'error')
+          })
+          .then(json => {
+            let newid = json["newid"]
+            if (newid) {
+              window.open(json["newid"] + "/browse", '_self');
+            } else {
+              sweetAlert('Oops', E_SERVER_ERROR, 'error')
+            }
+          })
```
jschwartzentruber: Is `.catch()` needed here in case we get an http error?

CovReporter/CovReporter.py
```diff
@@ -25,17 +25,23 @@
...
 __all__ = []
 __version__ = 0.1
 __date__ = '2017-07-10'
 __updated__ = '2017-07-10'
 
-# Debugging variables to keep track of GCOV errors
-null_coverable_count = 0
-length_mismatch_count = 0
-coverable_mismatch_count = 0
+# These variables are mainly for debugging purposes. We count the number
```
jschwartzentruber: Could this be part of the `CovReporter` instance rather than global?

server/covmanager/templates/collections/index.html
choller: Afaik, it should not be because fetch() does not raise on an HTTP error but instead returns the response but response.ok won't evaluate to true.

PR #497, opened 2018-07-25
Don't crash if $pc is not readable

gsingh93:
Sometimes the program counter points to a non-readable memory address after a crash. When we run the `gdb_cmds.txt` file in funfuzz, commands like `x/8i $pc` will fail. I'm not sure if this is the right solution, but I just check for the text 'Error in sourced command file' in the `crashInstruction` variable and set the instruction to `None` so it's ignored. I tested this locally and it seems to work.

### DIFF
--- a/FTB/Signatures/CrashInfo.py
+++ b/FTB/Signatures/CrashInfo.py
@@ -948,6 +948,8 @@ class GDBCrashInfo(CrashInfo):
         if self.crashInstruction is not None:
             # Remove any leading/trailing whitespaces
             self.crashInstruction = self.crashInstruction.strip()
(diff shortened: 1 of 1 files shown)


gsingh93:
@choller thoughts on this?

gsingh93:
ping @choller

nth10sd:
@jschwartzentruber may probably be a better person for now?

jschwartzentruber:
@gsingh93 Thanks for fixing this. Do you have an example log so we can add a test? I worry that `Error in sourced command file` is too non-specific. Is that all we get from GDB?

PR #500, opened 2018-08-03
Centralize AWS credentials into settings.py.

jschwartzentruber:
These are optional, so if not set, boto will look in the usual places.


jschwartzentruber:
It should be the same as if a PoolConfiguration is missing them now. The boto connection will raise and we'll get an error on the InstancePool.

PR #502, opened 2018-08-14
Allow non-ASCII in crash logs, when tested against a signature.

### LINKED ISSUE #501: [FTB] UnicodeDecodeError using signature.match with string match (non-PCRE)
Observed by @choller while fuzzing:
```
  Message: Error searching for signature: Traceback (most recent call last):
  File "/srv/repos/FuzzManager/Collector/Collector.py", line 602, in <module>
    sys.exit(main())
  File "/srv/repos/FuzzManager/Collector/Collector.py", line 531, in main
    (sig, metadata) = collector.search(crashInfo)
  File "/usr/local/lib/python2.7/dist-packages/FuzzManager-0.2.1-py2.7.egg/Reporter/Reporter.py", line 53, in decorator
    return wrapped(self, *args, **kwargs)
  File "/srv/repos/FuzzManager/Collector/Collector.py", line 179, in search
    if crashSig.matches(crashInfo):
  File "/usr/local/lib/python2.7/dist-packages/FuzzManager-0.2.1-py2.7.egg/FTB/Signatures/CrashSignature.py", line 104, in matches
    if not symptom.matches(crashInfo):
  File "/usr/local/lib/python2.7/dist-packages/FuzzManager-0.2.1-py2.7.egg/FTB/Signatures/Symptom.py", line 128, in matches
    if self.output.matches(line):
  File "/usr/local/lib/python2.7/dist-packages/FuzzManager-0.2.1-py2.7.egg/FTB/Signatures/Matchers.py", line 67, in matches
    return self.value in val
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 33: ordinal not in range(128)
```

jschwartzentruber:
If the input is not already unicode, try to interpret it as UTF-8. If there are errors, replace them with U+FFFD so we neither raise nor generate a false positive match.

Fixes #501.

### DIFF
--- a/FTB/Signatures/Matchers.py
+++ b/FTB/Signatures/Matchers.py
@@ -61,6 +61,11 @@ class StringMatch():
                     raise RuntimeError("Unknown match operator specified: %s" % matchType)
 
     def matches(self, val):
+        if isinstance(val, bytes):
+            # If the input is not already unicode, try to interpret it as UTF-8
+            # If there are errors, replace them with U+FFFD so we neither raise nor false positive.
+            val = val.decode("utf-8", errors="replace")
+
         if self.isPCRE:
             return self.compiledValue.search(val) is not None
         else:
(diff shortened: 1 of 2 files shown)


jschwartzentruber:
The fix is in StringMatch, so it would apply to all text symptoms (StackFrames, Instruction, Testcase).

PR #504, opened 2018-08-21
Make poolsize cores

jschwartzentruber:
This is a fairly large change, so I broke it up as logically as I could into commits.

- make spot price lookup periodic instead of on-demand (uses boto3 API)
- update Laniakea to latest API
- change pool size from instances to cores

I manually tested this by defining a pool with a size of 8 and allowing both c5.xlarge and c5.2xlarge. Depending on conditions, either 1x c5.2xlarge or 2x c5.xlarge was created.


README.md
```diff
@@ -68,7 +68,7 @@ requires the full repository to be checked out, not just the server directory.
 
 Server dependencies are listed in `server/requirements.txt`. You can use pip to install
 these dependencies using `pip install -r server/requirements.txt` and/or you can
-use your distribution's package management to fulfill them.
+use your distribution's package management to fulfill them. A redis server is also required.
```
choller: Would this also be required if the user doesn't want to use EC2SpotManager at all?

I think if we only use it there, then it should only be required if you actually use that module and in that case, I would like to explicitly state that here. Maybe also mention briefly how this can be installed with Ubuntu as an example.

I know it isn't part of this PR but eventually we should document the same for celery, because FuzzManager without celery is much less preferred by now.

server/ec2spotmanager/common/ec2.py
```diff
@@ -0,0 +1,17 @@
...
+    "eu-central-1",
+    "eu-west-1",
+    "eu-west-2",
+    "eu-west-3",
+    "sa-east-1",
+    "us-east-1",
+    "us-east-2",
+    "us-west-1",
+    "us-west-2"
+)
```
choller: Would it make sense to pull this information from our pool configurations instead? Otherwise, we will soon run into the problem that we have to update this file whenever a region is added or removed. Querying the database for all regions in use should be fairly quick.

README.md
jschwartzentruber: Sure. It will still be listed in the `requirements.txt`, but you're right, we're only using this so far in EC2SpotManager, and other modules would work without it.

As I mentioned offline, it also works as a celery message broker. We could switch the default celeryconf.py to use `redis://` instead of `ampq://` which is a much lighter dependency for local server use (redis-server is 168k installed, and runtime is C vs erlang).

server/ec2spotmanager/common/ec2.py
```diff
@@ -1,3 +1,148 @@
+# data exported from: https://ec2instances.info/ at 2018-08-16 17:54:16 UTC .. see disclaimers
+
+import collections
+
+
+InstanceType = collections.namedtuple("InstanceType", ("api_name", "vCPUs"))
+
+
+INSTANCE_TYPES = (
```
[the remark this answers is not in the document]
jschwartzentruber: I also forgot, this is used in the data migration to convert instance count to core count. I think it's better if that works without making any external requests.

PR #508, opened 2018-09-03
Add Report Configurations and Summary Functionality

choller:
Detailed changelog below:

Concept and prototypes for dynamic report configurations
Do basic cleanup operations in Collections and ReportConfiguration
Add ReportSummary model and extend ReportConfiguration
Add ReportSummary API, template and task
Squash report-related migrations into one

Switch the filtering directives language to glob-like instead of regex.
Shell style globs (* -> /.*/, ? -> /./) work only on a path segment
(file or directory name).
Rsync style ** is also supported for recursive matching, but only as a
complete path segment (ie. **/foo or foo/**, foo/**/bar, **/foo/**,
but NOT foo**).
Refactor to make it easier to see the algorithm used for glob filtering.

Add django-livesync for hot reloading
Improve the modal dialog UI
Add bootstrap-vue for having vue functionality in UI components
Add 'if debug' for Vue.options.debug
Add d3 dependency for graphs
Add Report Summary functionality to main page
Allow filtering by level on summary graph
Replace the dropdown with an ugly modal, sort rc list by description
Line up the title of the modal to the same line as the close button
Add local copies of d3 and vue-chartjs and include license files
Factor out cookie csrf code and switch d3/vue-chartjs to local files
Add local storage support for currently selected rc
Invalidate rc in local storage if it belongs to a deleted rc
Allow comments and empty lines in report configuration directives
Increase graph size and try to be (a little) smarter in min capping
Block search if cetain key modifiers are active (e.g. allow CTRL+C)
Make treemap capping relative to root, use ratio 1


choller:
I squashed this PR because I was in a hurry and didn't have time to sort the multiple dozen of commits.

Backend rewrite for Report Configuration Include/Exclude directives (to use glob style) was done by @jschwartzentruber 

Also @posidron helped a lot with frontend vue.js work like modal dialogs, CSS and better development process with django-livesync.

Thank you both!

PR #514, opened 2018-10-10
Add high-level permissions for viewing CovManager/CrashManager/EC2SpotManager.

### LINKED ISSUE #513: Should have ACLs for controlling access to FuzzManager modules
We should have some access controls for disabling access to FuzzManager modules like EC2SpotManager by user. Currently we have the restricted user flag which we have abused to get some control (disabling access to EC2SpotManager, restricting access to CrashManager), but this is already very limited and can't be extended further.

jschwartzentruber:
This adds `view_covmanager`, `view_crashmanager`, and `view_ec2spotmanager` permissions to hide entire portions of the FuzzManager server from unauthorized users.

The data migration gives all pre-existing users access to everything (as they had before), but any new users will have access to CovManager only (defined in `settings.py`).

These are only enforced in the web interface, not the REST API (yet).

Fixes #513


jschwartzentruber:
I went through each endpoint behind `/rest/`. The only two which allow SessionAuthentication are the signatures.zip download (which is denied for restricted users), and testcase download (which enforces tool view for restricted users). For CrashManager and EC2SpotManager, all other endpoints require TokenAuthentication.

To extend this to DRF, we would create a custom permission class (https://www.django-rest-framework.org/api-guide/permissions/), and use that instead of `IsAuthenticated`. For now the combination of this and `USERS_RESTRICTED_BY_DEFAULT` should be sufficient.

jschwartzentruber:
I went ahead and fixed it for DRF, so the REST interface has the same permission checks.

Issue #198, opened 2016-04-22
[FTB] Back/forward slash confusion on Windows paths


nth10sd:
Filing first before I forget how to describe. FuzzManager creates suggested values for the value of the stderr output correctly for directories with only forward slashes "/".

It is unsuccessful for those involving a mix of back and forward slashes which are sometimes generated on Windows: https://fuzzmanager.fuzzing.mozilla.org/crashmanager/signatures/new/?crashid=298057

Creating a bucket results in a suggestion of:

"/Assertion failure: !lookup\(l\)\.found\(\), at c:\\Users\\Administrator\\shell\-cache\\js\-dbg\-[0-9]{2,}\-dm\-windows\-[0-9]{2,}f0fa[0-9]{2,}c\\objdir\-js\\dist\\include\\js/HashTable\.h:[0-9]+/"

(note the mix of back & forward slashes above) instead of:

"/Assertion failure: !lookup\(l\)\.found\(\), at ([a-zA-Z]:)?/.+/HashTable\.h:[0-9]+/"

nth10sd:
Ref: https://fuzzmanager.fuzzing.mozilla.org/crashmanager/signatures/2016/

I found that changing the bucket's stderr output value to:
```
"/Assertion failure: !lookup\\(l\\)\\.found\\(\\), at ([a-zA-Z]:)?(/|\\\\).+/HashTable\\.h:[0-9]+/"
```
note the: ```(/|\\\\)``` portion, seems to allow for both back and forward slashes.


Also note that changing the above portion to ```(/|\\)``` then clicking Preview will cause:
```
error at /crashmanager/signatures/2016/edit/

unbalanced parenthesis

Request Method: 	POST
Request URL: 	https://fuzzmanager.fuzzing.mozilla.org/crashmanager/signatures/2016/edit/
Django Version: 	<masked>
Exception Type: 	error
```

nth10sd:
Traceback:
File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  111.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/decorators.py" in _wrapped_view
  22.                 return view_func(request, *args, **kwargs)
File "/srv/fuzzmanager.fuzzing.mozilla.org/FuzzManager/server/crashmanager/views.py" in editSignature
[...]

PR #516, opened 2018-10-23
[FTB] Support backslash in assertion message paths

jschwartzentruber:
This approach does the following:

- replace all backslashes in the assertion message with forward slashes
- create assertion pattern as normal (using forward slashes)
- restores backslashes that were not replaced by a pattern
- replaces the path pattern with one that accepts either back/forward-slash

I also removed escaping of `-` since this is not a special character except in a character range (which is already escaped).

Updated tests to match, and altered tests to actually try the resulting pattern instead of just compiling it.

I didn't remember about #198 before-hand. I guess this is an alternate approach to what was discussed there (normalizing input in Collector).

### DIFF
--- a/FTB/AssertionHelper.py
+++ b/FTB/AssertionHelper.py
@@ -247,7 +247,7 @@ def escapePattern(msg):
 
     escapedStr = msg
 
-    activeChars = ["\\", "[", "]", "{", "}", "(", ")", "*", "+", "-", "?", "^", "$", ".", "|"]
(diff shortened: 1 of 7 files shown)


jschwartzentruber:
Remaining issues with this approach:

* if you have a pattern matching a literal `\` and a path in the same line, we won't be able to match both in a windows crash
* performance hit of trying string match twice for windows crashes if the pattern contains `/`

PR #523, opened 2018-11-22
[FTB] Fix some flake8 warnings and nits

tysmith:
This will remove some flake8 warnings due to incorrect escaping. It also includes some other nits I came across that are easy to fix and may offer some performance benefits (and make things a bit more pythonic). No functional changes are intended.

### DIFF
--- a/FTB/AssertionHelper.py
+++ b/FTB/AssertionHelper.py
@@ -302,9 +302,10 @@ def escapePattern(msg):
 
     escapedStr = msg
 
-    activeChars = ["\\", "[", "]", "{", "}", "(", ")", "*", "+", "?", "^", "$", ".", "|"]
+    activeChars = ("\\", "[", "]", "{", "}", "(", ")", "*", "+", "?", "^", "$", ".", "|")
 
     for activeChar in activeChars:
-        escapedStr = escapedStr.replace(activeChar, "\\" + activeChar)
+        if activeChar in escapedStr:
+            escapedStr = escapedStr.replace(activeChar, "\\" + activeChar)
 
     return escapedStr
--- a/FTB/Signatures/CrashInfo.py
+++ b/FTB/Signatures/CrashInfo.py
@@ -480,7 +480,7 @@ class CrashInfo(object):
                 frame = frame[:idx]
 
         if "lambda" in frame:
-            frame = re.sub("<lambda at .+?:\d+:\d+>", "", frame)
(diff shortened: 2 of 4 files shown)


FTB/AssertionHelper.py
```diff
@@ -297,9 +297,10 @@ def escapePattern(msg):
 
     escapedStr = msg
 
-    activeChars = ["\\", "[", "]", "{", "}", "(", ")", "*", "+", "?", "^", "$", ".", "|"]
+    activeChars = ("\\", "[", "]", "{", "}", "(", ")", "*", "+", "?", "^", "$", ".", "|")
 
     for activeChar in activeChars:
-        escapedStr = escapedStr.replace(activeChar, "\\" + activeChar)
+        if activeChar in escapedStr:
```
choller: Is this really faster? replace is not doing a regex match and adding the additional if scans each string twice essentially instead of doing it once. I would assume that not changing the string at all is optimized and does not actually clone the string when assigning?

FTB/Signatures/CrashInfo.py
```diff
@@ -1041,10 +1041,10 @@ def isDerefOp(op):
             return "(" in op and ")" in op
 
         def calculateDerefOpAddress(derefOp):
-            match = re.match("\*?((?:\\-?0x[0-9a-f]+)?)\\(%([a-z0-9]+)\\)", derefOp)
+            match = re.match("\\*?((?:\\-?0x[0-9a-f]+)?)\\(%([a-z0-9]+)\\)", derefOp)
```
choller: Were these regexes broken before? We should have plenty of test coverage on these. Or does Python implicitly do the right thing here and this doesn't change semantics at all?

FTB/AssertionHelper.py
tysmith: Yep it is faster with one exception Python versions <2.7.15 in which case it is only faster on strings that are less than 200 characters. On newer Python versions I've seen speed ups of up to 6x.
So I guess we need to take in to account the Python version we run and the average string length that this function is used on? I put a bench-marking script on the server if you want to sanity check it.

At the very least a new string is created each iteration by `"\\" + activeChar`

choller:
I assume this PR is still current/required?

tysmith:
Correct. Feel free to merge or let me know if you'd like me to make any changes.

choller:
Merged, thanks!

PR #524, opened 2018-11-22
Broaden version requirements and upgrade some dependencies

choller:


choller:
Yes, livesync is a development-only package that can be used to automatically refresh browser pages when something in the backend or static files change. It is nice to have for developing, but not used in production. I disabled it because it has been unmaintained for quite a while and was pulling in old/incompatible dependencies like tornado 4.x, colliding with some requirements from jupyter.

PR #532, opened 2019-03-26
[FTB] Add support for Valgrind reports

tysmith:


FTB/Signatures/CrashInfo.py
```diff
@@ -232,6 +232,9 @@ def fromRawCrashData(stdout, stderr, configuration, auxCrashData=None, cacheObje
             elif minidumpFirstDetected and minidumpSecondString in line:
                 result = MinidumpCrashInfo(stdout, stderr, configuration, auxCrashData)
                 break
+            elif line.startswith("==") and re.match(ValgrindCrashInfo.MSG_REGEX, line):
```
choller: Would it make sense here to instead just check for the header of the Valgrind output itself, e.g. look for "== Using Valgrind" or "== Memcheck, ..."? The purpose of that code in fromRawCrashData is solely to decide somewhat reliably what crash info it is dealing with. For Valgrind traces (and potentially some other formats which output the ==PID== prefix), this will increase load because a lot of lines will have to be matched for that regex before finally finding some matching error message.

FTB/Signatures/CrashInfo.py
```diff
@@ -1677,3 +1680,85 @@ def createShortSignature(self):
...
+                continue
+            lineInfo = re.match(stackPattern, traceLine)
+            if lineInfo is None:
+                if self.backtrace:
+                    # at this point we should have the full trace
+                    # check if address info is available
+                    addr = re.match(r"^==\d+==\s+Address\s(?P<addr>0x[0-9A-Fa-f]+)\s", traceLine)
+                    if addr:
+                        self.crashAddress = int(addr.group("addr"), 16)
+                    # done parsing
```
choller: If I understand this code correctly (and I may well not :D), this ends the parsing after the first non-stack line has been found. So in case of reports where you have two stacks (e.g. the invalid free example in valgrind-if-01.txt), the second stack is not included? I believe for ASan we do include that stack and just append it to the first, so it is possible to match on that second stack as well. Can you maybe check that? We should try to keep that somewhat consistent, although I agree the stack concat in that particular case is not the best solution.

FTB/Signatures/CrashInfo.py
tysmith: "== Using Valgrind" or "== Memcheck, ..." are not always in the log. For example if Valgrind is run with -q you will only see the summary I have in most of the test cases. The check for "==" is in place to prevent running the regex more than necessary (when we can avoid it). If other tools using the "==PID==" prefix are supported we should detect them before scanning the entire log. Worst case is if we have a large log were all lines start with  "==" and no reports are detected, which I think could happen but will be fairly uncommon

FTB/Signatures/CrashInfo.py
tysmith: The functionality we should aim for is concatenating the use, alloc and free stacks if available, correct?

tysmith: All available stacks in a single report (not the entire log of course) will now be added to the backtrace.

FTB/Signatures/CrashInfo.py
choller: I see, thanks. We should make sure that we run with -q then to avoid scanning whole logs with regex. Because if we had to scan through that whole Valgrind header and various other output with regex on every submit before finding the first report, this could considerably affect the server load when receiving lots of these logs. If we use -q though and there is really only the report itself, then that should be fine.

PR #553, opened 2019-05-21
Add support for GCE as a cloud provider.

jschwartzentruber:
This implementation continues to use laniakea as its backend. This provider is different in that it uses docker images instead of userdata, and it uses preemptible VMs since there is no spot market.

Since Userdata Macros are usually used for environment, I made a checkbox to use them as GCE environment variables also. I also thought about doing this as "Common Environment", "EC2 Userdata Macros", and "GCE Environment", with EC2 and GCE both inheriting common environment automatically.

UI:

image
image


server/ec2spotmanager/CloudProvider/GCECloudProvider.py
```diff
@@ -0,0 +1,376 @@
...
+
+    def __delitem__(self, key):
+        return super(_LowercaseDict, self).__delitem__(key.lower())
+
+
+class GCECloudProvider(CloudProvider):
+    '''Implementation of CloudProvider interface for GCE.
+    '''
+    NODE_STATE_MAP = {
+        "PROVISIONING": INSTANCE_STATE["pending"],
```
choller: We could map this to INSTANCE_STATE["requested"] as well, right? Or does PROVISIONING mean that the request is already fulfilled?

server/ec2spotmanager/CloudProvider/GCECloudProvider.py
```diff
@@ -0,0 +1,376 @@
...
+            assert len(expr["tieredRates"]) == 1, expr["tieredRates"]
+            rate = expr["tieredRates"][0]
+            unit = int(rate["unitPrice"]["units"])
+            nanos = rate["unitPrice"]["nanos"]
+            return ((unit + float(nanos) / 1e9), expr["usageUnit"])
+
+        zones = {}
+        service_id = "6F81-5844-456A"  # Compute Engine
+        params = {"key": settings.GCE_API_KEY}
+        data = requests.get("https://cloudbilling.googleapis.com/v1/services/" + service_id + "/skus",
```
choller: Maybe make a variable for the URL and reuse that down below so we don't duplicate the URL

server/ec2spotmanager/CloudProvider/GCECloudProvider.py
```diff
@@ -0,0 +1,376 @@
...
+
+                    for region in sku["serviceRegions"]:
+                        zonedata = zones.setdefault(region, {})
+                        # TODO: the friendly description describes the zone, but I have no way of mapping
+                        #       that to an API zone (eg. Tokyo & Japan -> asia-northeast1-a & asia-northeast1-b
+                        zonedata.setdefault(what, []).append(price)
+
+            if data["nextPageToken"]:
+                params["pageToken"] = data["nextPageToken"]
+                data = requests.get("https://cloudbilling.googleapis.com/v1/services/" + service_id + "/skus",
```
choller: Duplicated URL here.

server/ec2spotmanager/CloudProvider/GCECloudProvider.py
jschwartzentruber: Probably yes. That's a translation from libcloud here: https://github.com/apache/libcloud/blob/trunk/libcloud/compute/drivers/gce.py#L1728

The point is moot though, since we only see Nodes returned once they are in the running state. I opened https://github.com/MozillaSecurity/laniakea/issues/41 about that.

jschwartzentruber: To be clear though, I will change that to "requested" ... I think you're right, it's the better representation in our model.

server/ec2spotmanager/CloudProvider/GCECloudProvider.py
jschwartzentruber: Thanks, I will.

jschwartzentruber:
> So your question about the environment variables is essentially about GCE being able to inherit EC2 userdata macros directly as environment, vs. having a "common environment" that both inherit?

Yes. Do you think the way I've done it is OK, or should I add a common environment? It's weird because EC2 doesn't directly have "environment", but macros have the same role (and we usually turn them into environment in userdata). My thinking with doing it this way is that it makes migration really easy, but that doesn't make it right.

PR #554, opened 2019-05-23
Coverage Reports

choller:
This is a first draft for the "Report" model in coverage, including new UI.


server/covmanager/templates/collections/report.html
```diff
@@ -17,76 +17,144 @@
...
     },
   },
   computed: {
-    ordered_collections: function () {
-      return _.orderBy(this.collections, [this.sortKey], [this.reverse ? 'desc' : 'asc'])
+    ordered_reports: function () {
+      if (this.reports) {
+        function formatMonthly(d) {
+          let months = [
+            "January", "Feburary", "March", "April", "May", "June",
```
jschwartzentruber: typo: February

server/covmanager/templates/collections/report.html
```diff
@@ -138,24 +210,67 @@
...
-          target.toggleClass("collection-selected", false)
+        self.selected_reports.splice(idx, 1)
       }
     },
+    publish: function (report) {
+      let url = URLS.reports_api + "update/" + report.id + "/"
+      let data = { "public" : !report.public }
+      this.update(url, data)
+    },
+    promote: function (report) {
```
jschwartzentruber: Does this refresh the view when you promote/demote, or just update the affected row?

ie. If you accidentally promote weekly->monthly, you need to click again twice to reset back to weekly. Is this going to refresh the page between each click?

server/covmanager/views.py
```diff
@@ -588,30 +588,54 @@ class CollectionViewSet(mixins.CreateModelMixin,
...
+        # Return early on empty queryset
+        if not queryset:
+            return queryset
 
-    report_filters = getattr(settings, 'COV_REPORTS_FILTER', ["Report"])
-    report_filters = [Q(description__contains=v) for v in report_filters]
-    filter = report_filters.pop()
+        can_see_unpublished = False
+        if request.user and request.user.is_authenticated:
+            can_see_unpublished = request.user.has_perm('crashmanager.view_crashmanager')
```
jschwartzentruber: We should create a separate permission for this in the future.

server/covmanager/views.py
```diff
@@ -588,30 +588,54 @@ class CollectionViewSet(mixins.CreateModelMixin,
...
 
-    for report_filter in report_filters:
-        filter |= report_filter
+        # We allow users to see unpublished reports, if they ask for it
+        # and also have the permissions to do so.
+        unpublished_requested = request.method != 'GET' or "unpublished" in request.GET
 
-    queryset = Collection.objects.filter(filter)
+        # TODO: Right now we use the view_crashmanager permission because that
```
jschwartzentruber: Please open an issue for this so we don't forget.

server/covmanager/templates/collections/report.html
choller: The data is loaded in the background and the view updated, so this is a data refresh but not a full page refresh. This is because the `update` function calls `fetch`.

server/covmanager/views.py
choller: Yes, I left a comment for that in another place where I used the crashmanager permission.

server/covmanager/views.py
choller: Filed #555. Note that this does not just affect Reports. Right now, `ReportConfigurations` can also be edited by anyone. We probably want to restrict that as well.

server/covmanager/serializers.py
```diff
@@ -150,3 +150,25 @@ def create(self, attrs):
 
         # Create our ReportConfiguration instance
         return super(ReportConfigurationSerializer, self).create(attrs)
+
+
+class ReportSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Report
+        fields = (
+            'id', 'data_created', 'public', 'coverage', 'is_monthly', 'is_quarterly'
```
jschwartzentruber: Is it worth adding a `validate()` method to ensure that we don't get `is_monthly=False` + `is_quarterly=True` ?

server/covmanager/templates/collections/report.html
jschwartzentruber: Perfect, thanks!

server/covmanager/templates/collections/report.html
```diff
@@ -138,24 +210,67 @@
...
-          target.toggleClass("collection-selected", true)
+        self.selected_reports.push(id)
       } else {
-        self.selected_collections.splice(idx, 1)
-        if (target)
-          target.toggleClass("collection-selected", false)
+        self.selected_reports.splice(idx, 1)
       }
     },
+    publish: function (report) {
```
jschwartzentruber: As discussed in IRC, `publish()` (and `promote()`) could have confirmation prompts before taking effect. Not a must, but it would make it a little safer from unnoticed, unintended clicks.

server/covmanager/migrations/0005_report.py
```diff
@@ -0,0 +1,28 @@
...
+    operations = [
+        migrations.CreateModel(
+            name='Report',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('data_created', models.DateTimeField(default=django.utils.timezone.now)),
+                ('public', models.BooleanField(default=False)),
+                ('is_monthly', models.BooleanField(default=False)),
+                ('is_quarterly', models.BooleanField(default=False)),
+                ('coverage', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='covmanager.Collection')),
```
jschwartzentruber: This line is too long by 1 character, which is why travis is failing.

server/covmanager/serializers.py
choller: I double-checked the code and all locations check `is_quarterly` first, none of it actually depends on them being both true. So I think we don't need to validate that. In fact, if that was a required property, I think we shouldn't use two bools in the first place in the model.

PR #574, opened 2019-08-19
[CovManager] Minor UI Improvements

pyoor:
This PR includes changes that:

- Replace the report configuration modal dialog with a simple drop-down
- Allow for case insensitive matching when filtering the tree list

### DIFF
--- a/server/covmanager/templates/collections/browse.html
+++ b/server/covmanager/templates/collections/browse.html
@@ -9,14 +9,15 @@
   <div class="panel-heading">
     <div>
       <i class="glyphicon glyphicon-tasks"></i> Coverage Collection Browser
-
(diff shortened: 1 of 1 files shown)


posidron:
Why not `overflow-y: scroll;` on `class="modal-body"` ?

pyoor:
I had initially tried that but ran into issues with the primary window being scrolled as opposed to the modal.  Further, the drop down is simpler and reduces the relevant code by quite a large margin.

posidron:
You might need to add `form-control` to the `select` class though, so bootstrap will recognize it. https://getbootstrap.com/docs/4.0/components/forms/?#form-controls But keep in mind that FM is using a 3.x version. https://getbootstrap.com/docs/3.4/css/#selects It saves code but it is also a different UI  approach .. but I need Lasik anyway. It needs to get fixed, so go for it I would say. :)

PR #578, opened 2019-10-24
Output error message for provider failures 

pyoor:
Currently, provider error messages are obscured by django request exceptions.  This PR adds a ProviderExceptions class and returns error messages containing the provider response.

### DIFF
--- a/server/crashmanager/Bugtracker/BugzillaProvider.py
+++ b/server/crashmanager/Bugtracker/BugzillaProvider.py
@@ -308,7 +308,13 @@ class BugzillaProvider(Provider):
         bz = BugzillaREST(self.hostname, username, password, api_key)
         ret = bz.createBug(**args)
         if ret.status_code != 200:
-            content = json.loads(ret.content)
+            try:
+                content = json.loads(ret.content)
+                if 'message' not in content:
+                    raise ValueError
+            except ValueError:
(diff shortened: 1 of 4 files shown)


server/crashmanager/Bugtracker/BugzillaProvider.py
```diff
@@ -302,14 +307,17 @@ def handlePOSTCreate(self, request, crashEntry):
         # Create the bug
         bz = BugzillaREST(self.hostname, username, password, api_key)
         ret = bz.createBug(**args)
-        if "id" not in ret:
-            raise RuntimeError("Failed to create bug: %s", ret)
+        if ret.status_code != 200:
+            content = json.loads(ret.content)
+            raise BugzillaProviderException(content['message'])
```
choller: We should probably try to handle more error cases here:

1) The resulting JSON might miss the `message`.
2) `ret.content` might not even parse as JSON (if the server has a hickup and returns nothing or e.g. the hardhat page)

In both cases, this would cause a 500 again. Instead we should probably wrap these cases in a `BugzillaProviderException` as well.

pyoor: Done

pyoor:
This is no longer necessary as the Bugzilla Provider now directly reports errors when creating bugs.

PR #596, opened 2020-02-14
Improve TSan support

choller:


FTB/Signatures/CrashInfo.py
```diff
@@ -1656,14 +1668,55 @@ def __init__(self, stdout, stderr, configuration, crashData=None):
...
 
+        if currentBacktrace is not None:
+            backtraces.append(currentBacktrace)
+
+        # For data races, TSan emits two stacks. If both stacks are intact,
+        # we should try to canonicalize their order to avoid creating multiple
+        # buckets for the same race.
+        if not brokenStack and isDataRace and len(backtraces) >= 2:
+            trace1 = backtraces.pop(0)
+            trace2 = backtraces.pop(0)
```
jschwartzentruber: This looks fine, however, could we not just sort them?

    backtraces = sorted(backtraces[:2]) + backtraces[2:]

It looks to me like this follows Python's list comparison semantics exactly.

PR #599, opened 2020-04-09
Remove numpy.

jschwartzentruber:
numpy is a heavyweight module to install just for the native type support which is easily emulated in Python.

I added doctests just to show that the operations are correct. (run with `python3 -m doctest FTB/Signatures/CrashInfo.py`, python2 will fail only because of `L` suffix on 64-bit results).

Casting everything to `uint32`/`uint64` before addition/multiplication means all math is in two's complement representation, where the final "cast" to `int32`/`int64` truncates the result to the desired range.

This could be slightly slower than using numpy, but I don't think this is in the critical path for signature matching.

### DIFF
--- a/FTB/Signatures/tests/test_CrashInfo.py
+++ b/FTB/Signatures/tests/test_CrashInfo.py
@@ -12,13 +12,13 @@ file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 @contact:    choller@mozilla.com
 '''
-from numpy import int64, uint64, int32, uint32
 import os
 
 from FTB.ProgramConfiguration import ProgramConfiguration
 from FTB.Signatures import RegisterHelper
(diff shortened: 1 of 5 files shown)


FTB/Signatures/tests/test_CrashInfo.py
```diff
@@ -1032,11 +1029,11 @@ def test_GDBParserTestCrashAddressSimple():
     # Note: The crash address here can also be 0xf7600000 because the double quadword
     # move can fail on the second 8 bytes if the source address is not 16-byte aligned
     assert (GDBCrashInfo.calculateCrashAddress("movdqu 0x40(%ecx),%xmm4", registerMap32) ==
-            int32(uint32(0xf75ffff8)))
```
choller, line 1035: Can you keep the original value here using `int32`/`uint32` as you defined it in `CrashInfo.py`? Turning this into a decimal makes the test harder to understand.

PR #602, opened 2020-04-27
Add TaskManager to track fuzzing tasks in TaskCluster.

jschwartzentruber:
* Add a read-only web interface to report fuzzing tasks running in TC
* TaskStatusReporter allows task status, identical to status data from EC2SpotManager.

N.B. Unlike EC2SpotManager, tasks are created externally, and therefore may try reporting status before we know about them. Reporters should work-around the fact that the report will 404 until
TaskManager finds the task and inserts it into the DB.

The TaskManager polling in `cron.py` uses the TaskManager Python client, which requires Python 3.


TaskStatusReporter/TaskStatusReporter.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python
+# encoding: utf-8
+'''
+EC2Reporter -- Simple EC2 status reporting tool for EC2SpotManager
```
choller, line 4: Ah, the smell of fresh copy pasta <3

server/requirements.txt
```diff
@@ -13,7 +13,7 @@ configparser>=3.5.0,<3.6; python_version == '2.7'
 coverage==4.5.2
 Django>=1.11.20,<3
 django-chartjs==1.3
-djangorestframework>=3.9.4,<3.10
```
choller, line 16: Have you checked why we added this requirement in the first place? I believe there was some sort of incompatibility with our code?

TaskStatusReporter/TaskStatusReporter.py
```diff
@@ -0,0 +1,153 @@
...
+__all__ = []
+__version__ = 0.1
+__date__ = '2014-10-01'
+__updated__ = '2014-10-01'
+
+
+class TaskStatusReporter(Reporter):
+    @functools.wraps(Reporter.__init__)
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('tool', 'N/A')  # tool is required by remote_checks, but unused by TaskStatusReporter
```
choller, line 45: Is there a particular reason for the decision to remove the tool association that we had for EC2?

server/taskmanager/views.py
```diff
@@ -0,0 +1,96 @@
...
+    queryset = Task.objects.all()
+    serializer_class = TaskSerializer
+    paginate_by_param = 'limit'
+    filter_backends = [
+        JsonQueryFilterBackend,
+        SimpleQueryFilterBackend,
+    ]
+
+    @action(detail=False, methods=['post'])
+    def update_status(self, request):
```
choller, line 69: Is this the entry point for machines updating their status? If so, does it require `SessionAuthentication` to be valid? I believe by having that, we allow CSRF to this API.

setup.cfg
```diff
@@ -3,7 +3,7 @@ universal = 1
 
 [metadata]
 name = FuzzManager
-version = 0.3.4
+version = 0.3.5
```
choller, line 6: Please be bold and make this 0.4 :D

TaskStatusReporter/TaskStatusReporter.py
jschwartzentruber: This is copy-paste from EC2Reporter. `Tool` is already not used by EC2SpotManager, only CovManager and CrashManager.

setup.cfg
jschwartzentruber: Will do

server/requirements.txt
jschwartzentruber: The functional changes in 3.10 and 3.11 were around schema support, which we don't use at all. 3.10 dropped support for Python 2.7, which is why I think we specified `<3.10`.

server/taskmanager/views.py
jschwartzentruber: Good catch! Hopefully this isn't too difficult to restrict within the same `ViewSet`. Maybe this is why we had separate views for each action in EC2SpotManager.

jschwartzentruber:
Sure, I will comment out TaskManager in `settings.py` by default and also the `taskcluster` dependency in `requirements.txt`.

PR #605, opened 2020-05-14
Fix taskcluster polling.

jschwartzentruber:


server/taskmanager/cron.py
```diff
@@ -71,17 +71,23 @@ def update_task_group(task_group_id):
                         "created": task["task"]["created"],
                         "state": run["state"],
```
choller, line 72: Hm, why is it `run["state"]` but `run.get(...)` for the others?

jschwartzentruber: If a run has been created, it has a state. The `started` and `resolved` fields are times that can be null depending on the state.

Issue #613, opened 2020-05-14
[FTB] Generic crashes on UBSan builds fail to parse


tysmith:
Non-UBSan specific crashes that are trigger on UBSan builds are not parsed properly.

```
==8555==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0xBAD (pc 0xBAD bp 0xBAD sp 0xBAD T28456)
==8555==The signal is caused by a READ memory access.
==8555==Hint: address points to the zero page.
    #0 0x7f070b805036 in foo() (somelib+0x12)
    #1 0x7f070b1f6961 in bar() (somelib.so+0x11)

UndefinedBehaviorSanitizer can not provide additional info.
==8555==ABORTING
```

PR #614, opened 2020-05-14
[FTB] Parse generic crashes found on UBSan builds

tysmith:
Use ASanCrashInfo() to parse generic crashes found by UBSan (non-ub crashes) since they are a subset of ASan errors.

This fixes issue #613.

### DIFF
--- a/FTB/Signatures/CrashInfo.py
+++ b/FTB/Signatures/CrashInfo.py
@@ -259,6 +259,7 @@ class CrashInfo(object):
         lsanString = "ERROR: LeakSanitizer:"
         tsanString = "WARNING: ThreadSanitizer:"
         ubsanString = ": runtime error: "
+        ubsanString2 = "ERROR: UndefinedBehaviorSanitizer"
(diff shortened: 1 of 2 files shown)


FTB/Signatures/CrashInfo.py
```diff
@@ -285,7 +286,7 @@ def fromRawCrashData(stdout, stderr, configuration, auxCrashData=None, cacheObje
             if ubsanString in line and re.match(ubsanRegex, line) is not None:
                 result = UBSanCrashInfo(stdout, stderr, configuration, auxCrashData)
                 break
-            elif asanString in line:
+            elif asanString in line or ubsanString2 in line:
                 result = ASanCrashInfo(stdout, stderr, configuration, auxCrashData)
```
jschwartzentruber, line 290: I don't like that we're handling ubsan crashes with the asan class when we have an ubsan crash. I realize the format is identical. I wonder if it's time for `xSanCrashInfo`

tysmith: OK I opened issue #615 to keep track of it.

PR #628, opened 2020-07-09
[EC2SpotManager] Fix excessive SQL load in pools index.

jschwartzentruber:
This optimizes the pool queries, and also caches all pool configs which are used in flatten.

### DIFF
--- a/server/ec2spotmanager/models.py
+++ b/server/ec2spotmanager/models.py
@@ -280,7 +280,9 @@ class PoolConfiguration(models.Model):
             return False
         tortoise = self
         hare = self._cache_parent(cache)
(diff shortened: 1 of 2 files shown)


server/ec2spotmanager/models.py
```diff
@@ -136,27 +136,29 @@ def __init__(self, *args, **kwargs):
...
+    def flatten(self, cache=None):
+        # cache is optionally a prefetched {config_id: config} dictionary used for parent lookups
+        if self.isCyclic(cache):
             raise RuntimeError("Attempted to flatten a cyclic configuration")
 
         self.deserializeFields()
 
-        # Start with an empty configuration
-        flat_parent_config = FlatObject({})
+        # If we are not the top-most confifugration, recursively call flatten
```
choller, line 146: nit: spelling

PR #631, opened 2020-07-23
Allow unknown task status, and better health reports.

jschwartzentruber:
Some projects report status once at boot, but taskmanager usually doesn't know about tasks in a pool until some time has passed (since it is polled). This allows reporting from unknown tasks, but with a short expiry in case the pool is never identified.

We now support non-continuous pools, eg. 2h/day or 24h/week. We should report these as "healthy" if they ran the requested time in the last cycle, even if they aren't running now.

This also adds significant testing to the taskcluster polling cron task.


server/taskmanager/serializers.py
```diff
@@ -11,13 +21,43 @@ class Meta:
...
+            query = (
+                Q(pool=instance) &
+                Q(started__isnull=False) &
+                (Q(resolved__isnull=True) | Q(resolved__gte=last_cycle_start))
+            )
+            for task in Task.objects.filter(query):
+                begin = max(task.started, last_cycle_start)
+                end = task.resolved or now
+                run_time += (end - begin).total_seconds()
+            run_ratio = run_time / ret["cycle_time"]
```
choller, line 52: Does this need a `float()` somewhere?

jschwartzentruber: Yes, it could. However the taskmanager module requires [fuzzing-tc,](https://github.com/MozillaSecurity/fuzzing-tc) which requires Python 3. In Python 3, `float()` would not have any effect. I'll add it for clarity.

PR #637, opened 2021-01-29
Improve the REST API

jschwartzentruber:
* implement same toolfilter logic in API as in UI
* allow session authentication
* add remaining fields to crashes API: `created` & `triagedOnce`
* add ordering filter to both crashes & buckets APIs
* make error handling more consistent (4xx instead of 5xx)

### DIFF
--- a/server/crashmanager/views.py
+++ b/server/crashmanager/views.py
@@ -1,6 +1,6 @@
 from collections import OrderedDict
 from datetime import timedelta
-from django.core.exceptions import SuspiciousOperation, PermissionDenied
+from django.core.exceptions import FieldError, SuspiciousOperation, PermissionDenied
 from django.core.paginator import Paginator, PageNotAnInteger, EmptyPage
 from django.db.models import F, Q
 from django.db.models.aggregates import Count, Min, Max
@@ -11,8 +11,10 @@ import functools
 import json
 import operator
 import os
-from rest_framework import filters, mixins, viewsets
+from rest_framework import mixins, viewsets
 from rest_framework.authentication import TokenAuthentication, SessionAuthentication
+from rest_framework.exceptions import MethodNotAllowed
+from rest_framework.filters import BaseFilterBackend, OrderingFilter
 from rest_framework.response import Response
 from rest_framework.views import APIView
 import six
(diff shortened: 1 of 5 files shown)


server/crashmanager/views.py
```diff
@@ -1263,23 +1296,28 @@ def partial_update(self, request, pk=None):
...
+    filter_backends = [ToolFilterSignaturesBackend, JsonQueryFilterBackend, BucketAnnotateFilterBackend, OrderingFilter]
 
     def retrieve(self, request, *args, **kwargs):
-        deny_restricted_users(request)
-        return super(BucketViewSet, self).retrieve(request, *args, **kwargs)
-
-    def list(self, request, *args, **kwargs):
-        deny_restricted_users(request)
-        return super(BucketViewSet, self).list(request, *args, **kwargs)
+        user = User.get_or_create_restricted(request.user)[0]
```
choller, line 1307: So restricted users here will get the same view as on the web api?

choller (approved):
lgtm! Thanks

server/crashmanager/views.py
jschwartzentruber: One big difference is that a request out-of-filter by a restricted user will result in a 404 instead of 403.  It's implemented using the same functions we use in web UI (`filter_crash_entries_by_toolfilter` & `filter_signatures_by_toolfilter`), but as DRF filter backends. So if an object is filtered out by toolfilter, it will just not be found and DRF will 404 automatically. I verified all this in tests.

|                                       |             Unrestricted user             | Restricted user  |
|---------------------------------------|-------------------------------------------|------------------|
|               /crashes/               |                 filtered                  |     filtered     |
|     /crashes/?ignore_toolfilter=1     |                    all                    |     filtered     |
|            /crashes/_{id}_/             |                    any                    | in-filter or 404 |
|            /crashes/_{id}_/?ignore_toolfilter=1             |                    any                    | in-filter or 404 |
|             /signatures/              |                 filtered                  |     filtered     |
|   /signatures/?ignore_toolfilter=1    |                    all                    |     filtered     |
|           /signatures/_{id}_/           | any (`size,quality` are filtered or `0,None`) |     in-filter or 404     |
| /signatures/_{id}_/?ignore_toolfilter=1 |                    any                    |     in-filter or 404     |

I documented this in the test files ([here](https://github.com/MozillaSecurity/FuzzManager/pull/637/files#diff-853e1c2cd2b57381fad277c5f95ea05c33418b9fd296c6d6a7bb39e7d4d9c207R27) and [here](https://github.com/MozillaSecurity/FuzzManager/pull/637/files#diff-89c7064d1e414c90e39f20f9b12c53bb53cd7bec06107005d6073f57f42c50d8R18)), because that's where I added tests to make sure each view is filtered properly.

PR #645, opened 2021-04-19
[libFuzzer] Add argument for performing post crash transformations

pyoor:
This PR adds a single argument for applying a post crash transformation to the testcase.  The `--transform` argument expects a path to a binary file that will be run with the testcase path as an argument.

### DIFF
--- a/misc/afl-libfuzzer/afl-libfuzzer-daemon.py
+++ b/misc/afl-libfuzzer/afl-libfuzzer-daemon.py
@@ -17,21 +17,11 @@ file, You can obtain one at http://mozilla.org/MPL/2.0/.
 # Ensure print() compatibility with Python 3
 from __future__ import print_function
 
-from Collector.Collector import Collector
-from FTB.ProgramConfiguration import ProgramConfiguration
-from FTB.Running.AutoRunner import AutoRunner
-from FTB.Signatures.CrashInfo import CrashInfo
-
-from S3Manager import S3Manager
-
 import argparse
 import collections
-from fasteners import InterProcessLock
 import os
-from pathlib import Path
 import re
 import shutil
-from six.moves import queue
 import stat
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


pyoor:
@choller - I just got around to pushing these changes.  This now passes an output directory to the transform script.

misc/afl-libfuzzer/afl-libfuzzer-daemon.py
```diff
@@ -621,6 +635,39 @@ def test_binary_asan(bin_path):
...
+
+        if len(os.listdir(output_path)) == 0:
+            raise Exception("Transformation script did not generate any files.  Aborting...")
+
+        archive_path = testcase_path + ".zip"
+        with ZipFile(archive_path, 'w') as archive:
+            archive.write(testcase_path)
+            for root, dirs, files in os.walk(output_path):
+                for file in files:
+                    archive.write(os.path.join(root, file))
```
choller, line 666: I'm not sure this is correct. With `output_path` being a full path (`/tmp/something`), that will the full path structure into the zip, won't it? Can you double check that this just includes one relative directory and in general the zip looks like you would expect it to?

misc/afl-libfuzzer/afl-libfuzzer-daemon.py
```diff
@@ -555,6 +561,14 @@ def scan_crashes(base_dir, collector, cmdline_path=None, env_path=None, test_pat
             if base_env:
                 env = dict(base_env)
 
+            submission = crash_file
+            if transform:
+                try:
+                    submission = apply_transform(transform, crash_file)
+                except Exception as e:
```
choller, line 568: I think we shouldn't die on exceptions here without submitting at least something that might make it possible to debug why the transform failed. We should probably send at least the untransformed testcase before dying.

choller (changes requested):

misc/afl-libfuzzer/afl-libfuzzer-daemon.py
pyoor: You are correct.  I've updated this to use pathlib and only include directories relative to the output path.

misc/afl-libfuzzer/afl-libfuzzer-daemon.py
pyoor: I removed the return statement from these failed attempts.  Instead of exiting, we warn and the original testcase will still be submitted.

choller (approved):

PR #651, opened 2021-04-29
Add missing database migrations

La0:
While working on the docker setup, I noticed some database migrations are missing.

Three migrations were automatically generated by Django:

```console
$ python server/manage.py makemigrations
Migrations for 'covmanager':
  server/covmanager/migrations/0006_auto_20210429_0908.py
    - Alter field file on collectionfile
Migrations for 'crashmanager':
  server/crashmanager/migrations/0003_auto_20210429_0908.py
    - Alter field test on testcase
Migrations for 'ec2spotmanager':
  server/ec2spotmanager/migrations/0002_auto_20210429_0908.py
    - Alter field ec2_userdata_file on poolconfiguration
    - Alter field type on poolstatusentry

```


La0:
CI is failing due to #652

server/covmanager/migrations/0006_auto_20210429_0908.py
```diff
@@ -0,0 +1,23 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='collectionfile',
+            name='file',
+            field=models.FileField(
+                max_length=255,
+                storage=django.core.files.storage.FileSystemStorage(location=None),
+                upload_to='coverage'
```
jschwartzentruber, line 20: The only change I can see from [0002_increase_collection_filename_length.py](/MozillaSecurity/FuzzManager/blob/master/server/covmanager/migrations/0002_increase_collection_filename_length.py) is `upload_to` changes from `bytes` to `str`. 👍🏻

server/crashmanager/migrations/0003_auto_20210429_0908.py
```diff
@@ -0,0 +1,24 @@
...
+        ('crashmanager', '0002_add_taskmanager'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='testcase',
+            name='test',
+            field=models.FileField(
+                storage=django.core.files.storage.FileSystemStorage(
+                    location='/home/lao/dev/mozilla/FuzzManager/server'
```
jschwartzentruber, line 19: The current source refers to `/home/decoder` ... so changing to `/home/lao` isn't better, just different. I don't know what the solution is to this problem. Can we reference a setting instead?

server/ec2spotmanager/migrations/0002_auto_20210429_0908.py
```diff
@@ -0,0 +1,29 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='poolconfiguration',
+            name='ec2_userdata_file',
+            field=models.FileField(
+                blank=True,
+                null=True,
+                storage=ec2spotmanager.models.OverwritingStorage(location='/home/lao/dev/mozilla/FuzzManager/server'),
```
jschwartzentruber, line 20: Ditto

server/ec2spotmanager/migrations/0002_auto_20210429_0908.py
```diff
@@ -0,0 +1,29 @@
...
+                upload_to=ec2spotmanager.models.get_storage_path
+            ),
+        ),
+        migrations.AlterField(
+            model_name='poolstatusentry',
+            name='type',
+            field=models.IntegerField(),
```
jschwartzentruber, line 27: The old field had `default=0` (which we have never specified in the model). 👍🏻

jschwartzentruber (commented):
We've always had to ignore the location changes when generating migrations. Can we use settings to make these not change depending on development environment?

server/crashmanager/migrations/0003_auto_20210429_0908.py
La0: I tried using [a callable to define storage](https://docs.djangoproject.com/en/dev/topics/files/#using-a-callable) but it's only available in Django 3.1

So i'm using  `settings.BASE_DIR` in the migration.

La0:
After rebasing on master, the CI is now OK.

jschwartzentruber (approved):

PR #664, opened 2021-05-12
Remove support for Python 3.5

### LINKED ISSUE #663: Remove Python 3.5 support
Simply remove the tox 3.5 configuration and the CI calls for this environment.

This would allow us to use f-strings in the codebase. Also python 3.5 has no security support [since September 2020](https://endoflife.date/python)

La0:
Closes #663

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
-envlist = py{35,36,37,38}
-minversion = 3.2
+envlist = py{36,37,38}
+minversion = 3.6
 tox_pip_extensions_ext_venv_update = true
 skip_missing_interpreters = true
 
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
-envlist = py{35,36,37,38}
-minversion = 3.2
+envlist = py{36,37,38}
+minversion = 3.6
```
jschwartzentruber, line 3: This refers to Tox version, not Python version.

jschwartzentruber (changes requested):
Looks good besides the Tox `minversion` change. I'm not sure that the existing v3.2 makes sense, but in any case, it is unrelated to dropping Python 3.5.

tox.ini
La0: TIL

jschwartzentruber (approved):

PR #673, opened 2021-05-14
Fix an inconsistent behavior on watched buckets filtering

### LINKED ISSUE #672: [CrashManager] Inconsistent behavior on watched buckets filtering
While refactoring the crash list, I noticed an inconsistent behavior that occurs when counting new crashes added on a watched bucket.

In `watchedSignatures` view (`server/crashmanager/views.py`) buckets are automatically filtered with `crashentry__tool__in=user.defaultToolsFilter.all()` even if the user doesn't have any `defaultToolsFilter` , it will result in an empty list of new crashes.

On the other hand, `ToolFilterSignaturesBackend`  uses `filter_signatures_by_toolfilter` function that will return all signatures if the user doesn't have any `defaultToolsFilter` .

The `watchedSignatures` view contains an explicit comment saying:
```
    # the join of Bucket+CrashEntry is a big one, and each filter() call on a related field adds a join
    # therefore this needs to be a single filter() call, otherwise we get duplicate crashentries in the result
    # this is why tool filtering is done manually here and not using filter_signatures_by_toolfilter()
```
 So I guess, the expected behavior is the one from `filter_signatures_by_toolfilter`  and that we should update `watchedSignatures` consequently to not filter out buckets when the user don't have some `defaultToolsFilter`.

EvaBardou:
Closes #672

### DIFF
--- a/server/crashmanager/views.py
+++ b/server/crashmanager/views.py
@@ -185,12 +185,17 @@ def watchedSignatures(request):
     # 3         blah        0           tr
     user = User.get_or_create_restricted(request.user)[0]
 
+    filters = {
+        'user': user,
+        'crashentry__gt': F('bucketwatch__lastCrash'),
+    }
     # the join of Bucket+CrashEntry is a big one, and each filter() call on a related field adds a join
     # therefore this needs to be a single filter() call, otherwise we get duplicate crashentries in the result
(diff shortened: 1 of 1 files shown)


jschwartzentruber (approved):
thanks!

PR #675, opened 2021-05-17
Simplify Bugzilla Provider

La0:
- Remove `Bugtracker.Provider`
- Rename module from `Bugtracker` to `bugzilla`
- Rename `Bugtracker.BugzillaRes` to `bugzilla.api`
- Rename `Bugtracker.BugzillaProvider` to `bugzilla.provider`
- Remove in model `Bugprovider` the field `classname` as it could only be `BugzillaProvider`
- Remove all usages of `Bugprovider.classname` in forms & views


La0:
After discussion with @choller it's not mergeable 

> it is a breaking change for someone who built an implementation on top of this

PR #679, opened 2021-05-20
Add a component to add/edit/remove a Provider API key in localStorage

### LINKED ISSUE #676: Store Bugzilla API token through JS localstorage
As we cannot store bugzilla API keys in the database (out of security concerns), we can still store the API token on the user's browser in order to make authenticated requests:
- the Bugzilla API key stays on the user computer and is never stored on the server
- a Vue.js form should be added in the user profile to allow him to set/unset the Bugzilla API key stored on the current browser context
- all FuzzManager API endpoints that use the `BugzillaProvider` should support an header `X-Bugzilla-API-Key` sent by the browser.
- only Javascript would easily get access to the token in `localStorage`, so we would really need to convert configuration loading and bug creation to that method (already planned)  
- only API key would be supported, username/password authentication would be removed

EvaBardou:
Closes #676

### DIFF
--- a/server/crashmanager/templates/usersettings.html
+++ b/server/crashmanager/templates/usersettings.html
@@ -1,51 +1,57 @@
 {% extends 'layouts/layout_base.html' %}
 
 {% block body_content %}
-<div class="panel panel-default">
-	<div class="panel-heading"><i class="glyphicon glyphicon-cog"></i> User Settings</div>
-	<div class="panel-body">
-		{% if msg %}<p><i>{{ msg }}</i></p>{% endif %}
-		<form action="{% url 'crashmanager:usersettings' %}" method="post">
-			{% csrf_token %}
-			Select the tools you would like to include in your default views:
-			{% for tool in tools %}
-			<div class="field">
-				<input type="checkbox" id="id_tool_{{ tool.name }}" name="tool_{{ tool.name }}" value="unassigned" {% if tool.checked %}checked{% endif %} />
-				<strong>{{ tool.name }}</strong>
-			</div>
-			{% endfor %}
-			<input type="hidden" name="changefilter" value="1" />
-			<input type="submit" name="submit" value="Save Tool Filters" class="btn btn-danger"/>
-		</form>
-		<hr/>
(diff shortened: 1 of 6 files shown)


server/frontend/src/components/ProviderKey.vue
```diff
@@ -0,0 +1,73 @@
...
+  data: () => ({
+    key: null,
+  }),
+  mounted() {
+    const storedKey = localStorage.getItem(this.localStorageKey);
+    if (storedKey) this.key = storedKey;
+  },
+  computed: {
+    localStorageKey() {
+      return "provider-" + this.providerId;
```
La0, line 51: Nit: use `provider-XX-api-key`

We may store other things later on, who knows

server/crashmanager/templates/usersettings.html
```diff
@@ -39,6 +39,13 @@
 			<input type="hidden" name="changetemplate" value="1" />
 			<input type="submit" name="submit" value="Save Default Provider/Template" class="btn btn-danger"/>
 		</form>
+		{% if providers %}
```
La0, line 42: I would create a second panel named `Bug Providers settings`, and move that code in the new panel

La0 (commented):
Looks good, I was able to create , view in dev console and remove an api key

server/frontend/src/components/ProviderKey.vue
```diff
@@ -0,0 +1,73 @@
...
+    },
+  },
+  methods: {
+    saveKey() {
+      localStorage.setItem(this.localStorageKey, this.key);
```
La0, line 56: It would be pretty nice to make an authenticated query towards the Bugzilla server (as you already have the hostname here) to check if the api key is valid.

You just need to make a query on [whoami](https://bmo.readthedocs.io/en/latest/api/core/v1/user.html#who-am-i) and display the real name of the user (or an error)

La0 (commented):
Bonus :gift:

La0:
That works nicely now, thanks !

pyoor (approved):

PR #684, opened 2021-05-26
Manage Bugzilla templates through Django forms

### LINKED ISSUE #680: Management of Bugzilla template through forms
Instead of managing BugzillaTemplate through the BugzillaProvider class, we could create  two forms (one for bug, the other for comments) in order to create and update those templates.
The current model implementation is not tied to any BugProvider, so it’s possible to manage them independently.

Benefits:
 - Simpler system, closer to Django standards
 - Error handling
 - No need for Vue.js, pure Django is OK here

First, the BugzillaTemplate model would get a new “mode” enum field (either bug or comment) to be able to display the desired mode.

Then we would be able to entirely re-structure the Django views & urls to manage templates, by leveraging the new Form.
Generic urls+views would be :
- list all the templates for a provider (bug & comment)
     - Url: `bugzilla/templates`

- edit bug & comment template through same url (they use the same view already)
     - Url: `bugzilla/templates/<template>`

- delete a template (bug & comment)
     - Url: `bugzilla/templates/<template>/delete`

 - create a bug template
   - Url: `bugzilla/templates/create-bug`

 - create a comment template
   - Url: `bugzilla/templates/create-comment`

All the current code related to templates management would be removed:
 - Provider methods `handlePOSTCreateEditTemplate` and `renderContextViewTemplate`
 - All existing HTML templates used to display & submit forms
 - All existing related Django views

As this part of the BugProvider is not easily extendable by 3rd party, this should not affect other developers (they would only have to remove some methods from their own Provider)

Finally, a new read-only API viewset would be created to allow Vue.js components to list and retrieve bugzilla templates.

EvaBardou:
Closes #680

This MR will load Bugzilla products & components through a Vue.js component in #628 (for now it's a manually editable field).

Important changes:
- Relevant Bug provider methods removed (from Bugzilla implementation and the base class)
- New views created using Django generic class-based views and Django forms, using [django-crispy-forms](https://django-crispy-forms.readthedocs.io/en/latest/) to cleanly render the forms using Bootstrap
- New read-only API endpoint to list and retrieve Bugzilla templates (will be used later in bug and comment publication )
- Unit tests to cover the views

You can access the templates list by clicking on the user menu, and then **Bugzilla templates** below **Settings**.

## List all templates (bug & comment)
Screenshot 2021-05-26 at 15-53-04 Screenshot

## Form for a bug template

Used on edition and creation.

:warning: This is where the Bugzilla products & components through Vue.js (from #628) will replace the two `<input>` fields. This was better to split in two MR (note from @La0 )

Screenshot 2021-05-26 at 15-53-14 Screenshot

## Form for a comment template

Used on edition and creation

Screenshot 2021-05-26 at 15-53-31 Screenshot

## Delete a template

Through a confirmation step

Screenshot 2021-05-26 at 15-53-46 Screenshot


La0 (approved):
LGTM :+1:

server/crashmanager/templates/bugzilla/list.html
```diff
@@ -0,0 +1,53 @@
...
+                    <a title="Next" href="?page={{ page_obj.next_page_number }}" data-placement="top" class="glyphicon glyphicon-chevron-right"></a>
+                {% endif %}
+            </span>
+        </div>
+        <table class="table table-condensed table-hover table-bordered table-db">
+            <thead>
+            <tr>
+                <th style="width: 20px;">ID</th>
+                <th style="width: 40px;">Name</th>
+                <th style="width: 30px;">Mode</th>
```
jschwartzentruber, line 31: Could we change `Mode` to `Type`? I think it is easier for a user to think that the template "has this type" instead of it "is in this mode". (only here where it is displayed, `.mode` in the model is fine)

jschwartzentruber (approved):
Looks good with one comment. Thanks for adding tests!

server/crashmanager/forms.py
```diff
@@ -0,0 +1,131 @@
...
+            'description': 'Bug description',
+            'security': 'This is a security bug',
+            'security_group': 'Security group',
+            'testcase_filename': 'Filename that will be used for the testcase',
+        }
+
+        widgets = {}
+        for field in fields:
+            if field not in ['description', 'attrs', 'security']:
+                widgets[field] = Textarea(attrs={'rows': 1})
```
pyoor, line 107: Just curious but is there a reason you're using textarea for all of these inputs?  It makes sense for `description` and possibly `attrs` but shouldn't the rest just be inputs?

server/crashmanager/templates/bugzilla/list.html
EvaBardou: Sure, no problem!

server/crashmanager/forms.py
EvaBardou: In `BugzillaTemplate` class all fields are declared as `models.TextField()` ones.

Django automatically converts them to `TextAreas` when using forms (see: [https://docs.djangoproject.com/en/3.1/ref/models/fields/#textfield](https://docs.djangoproject.com/en/3.1/ref/models/fields/#textfield)) so I kept this "convention".

But if you want, I could easily change those to display `TextInputs` instead.

jschwartzentruber: That was sloppy on our part in defining the template model... I think they probably should be `CharField(max_length=4095)`? If you'd rather specify `TextInput` instead, that's fine.

PR #687, opened 2021-05-28
Use v-pre to prevent Vue.js failure on special characters

### LINKED ISSUE #686: Vue.js failure on crash with special characters
Some crash fail to render through Vue.js with an error `invalid expression: invalid escape sequence in ...` in the console.

An example has been found by @jschwartzentruber : https://fuzzmanager.teklia.com/crashmanager/crashes/107/

I produced a dump of the db + crashes data.

EvaBardou:
Closes #686 

Since Vue.js is now available directly at the project root, it was trying to compile the displayed code from `rawStdout` attribute on `CrashEntry` 107.

By using [`v-pre`](https://vuejs.org/v2/api/#v-pre) attribute on HTML elements potentially rendering code, we prevent Vue.js from misinterpreting data to be rendered.

### DIFF
--- a/server/crashmanager/templates/crashes/view.html
+++ b/server/crashmanager/templates/crashes/view.html
@@ -62,7 +62,7 @@
 	    <span class="glyphicon glyphicon-eye-open"></span>
             </button>
 	    <div id="stdout" class="accordian-body collapse">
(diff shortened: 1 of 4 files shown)


jschwartzentruber (approved):
👍🏻

PR #688, opened 2021-05-28
Add back limited Taskmanager scraping

jschwartzentruber:
#649 added the ability to listen for mozillapulse updates rather than scraping Taskcluster all the time. This mostly works great, but we do miss updates occasionally. 

This adds scraping only for specific tasks that we already know about in the database and that are unresolved (pending/running).

Also fix a traceback seen when updating pools from fuzzing-tc-config.

### DIFF
--- a/server/taskmanager/cron.py
+++ b/server/taskmanager/cron.py
@@ -2,11 +2,36 @@
 from __future__ import unicode_literals
 from datetime import datetime
 from logging import getLogger
+from django.conf import settings
 from django.utils import timezone
 from celeryconf import app
 
 
 LOG = getLogger("taskmanager.cron")
+
+
+@app.task(ignore_result=True)
+def update_tasks():
+    import taskcluster
+    from .models import Task
+    from .tasks import update_task
+
+    queue_svc = taskcluster.Queue({"rootUrl": settings.TC_ROOT_URL})
+    now = datetime.now(timezone.utc)
+
+    # this should never be required, but we seem to miss Task events
+    # in mozilla pulse sometimes.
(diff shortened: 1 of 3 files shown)


server/taskmanager/cron.py
```diff
@@ -2,13 +2,38 @@
...
+
+    queue_svc = taskcluster.Queue({"rootUrl": settings.TC_ROOT_URL})
+    now = datetime.now(timezone.utc)
+
+    # this should never be required, but we seem to miss Task events
+    # in mozilla pulse sometimes.
+    # if we notice that a task is pending or running for longer than
+    # normal, try to update the task directly from taskcluster
+
+    for task_obj in Task.objects.filter(state__in=["pending", "running"]).select_related("pool"):
```
jschwartzentruber, line 27: I considered adding `or state is None` here (using Django `Q()`) which would update tasks that have status from `Taskreporter` but haven't been reported by `mozillapulse` yet.  This would mean tasks would still show up here if `mozillapulse` was down or not configured, but would be racey and unnecessary if `mozillapulse` is configured.

wyt?

pyoor: I think what you have is fine.  Though, and I have no idea where this should go, but maybe we should consider adding an alert in the UI that `mozillapulse` is down or not configured.

pyoor (approved):
Looks good to me.

server/taskmanager/cron.py
jschwartzentruber: I'll create an issue + label for notification ideas once the system is in place.

PR #689, opened 2021-05-28
[libfuzzer] Exit gracefully when attempting to merge an empty corpus

pyoor:
During the merge process, if the corpus doesn't exist in S3, we return a fatal error.  Instead, we should exit gracefully without attempting to merge.


jschwartzentruber (approved):

pyoor:
I'm not sure if this is actually correct.  Now that the corpus I was trying to load has been created I can't test this without creating a dummy corpus.  I'm going to close this for now.

PR #690, opened 2021-05-31
Vueify bugzilla product & component selection

### LINKED ISSUE #682: Vueify bugzilla product & component selection
A new Vue.js component should be created to allow a user picking a Bugzilla product and component.

It will be used on several pages:
- Bugzilla Bug template creation and edition
- Bugzilla Bug publication

It will display 3 select fields:
1. the list of Bugzilla providers, displaying their hostname
2. the list of products for that provider hostname
3. the list of components for the selected product.

First we'll need a new read-only API viewset to list available providers on `/rest/providers`. The viewset should simply exports all 3 fields: `classname`, `hostname` and `url_template`.

The vue.js component would fetch the providers when mounted, and list all the `classname=BugzillaProvider` ones. When a provider is selected it will :
1. look in localStorage if the key `bugzilla-products-{hostname}` is available
2. if it's available use it to display products
3. otherwise, load `https://{hostname}/latest/configuration` and store the components and products in localstorage (not the full payload)

When a product is selected, display the components from that product.

Finally, an extra button is added to refresh the data loaded from the bugzilla instance into localstorage. It's named "Refresh" and displayed next to the provider list when a provider is selected.

This code will replace the javascript implementation in `crashmanager/templates/bugzilla/create_form.html`.

## Configuration parsing and storage

The configuration for BMO is available at https://bugzilla.mozilla.org/latest/configuration (:warning: it's a laaaarge JSON).

We are only interested by a few keys in that file, so we can store a subset of data in `localStorage`:

- keep `version`
- iterate over `product` to build the list `products`
  - filter on `is_active=true`
  - build a new dict with key `name`, `id` and an empty array `components`
  - iterate over each `component`:
    - filter on `is_active=true`
    - only keep the name in the array

You should end up with the following structure:

```json
{
  "version": "20210517.1",
  "products": [
    {
      "id" : 1,
      "name" : "Taskcluster",
      "components" : [
         "General",
         "Operations and Service requests",
         ...
      ]
    },
    ....
  ],
}
```

EvaBardou:
Closes #682

### DIFF
--- a/server/crashmanager/forms.py
+++ b/server/crashmanager/forms.py
@@ -1,6 +1,6 @@
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import HTML, Layout, Div, Field, Submit
-from django.forms import ModelForm, Textarea, TextInput
+from django.forms import CharField, ModelForm, Textarea, TextInput
 
 from .models import BugzillaTemplate
 
@@ -18,6 +18,7 @@ class BugzillaTemplateBugForm(ModelForm):
             Field('product', wrapper_class='col-md-6'),
             Field('component', wrapper_class='col-md-6'),
         ),
+        HTML("""<productcomponentselect></productcomponentselect>"""),
         Row(
             Field('whiteboard', wrapper_class='col-md-6'),
             Field('keywords', wrapper_class='col-md-6'),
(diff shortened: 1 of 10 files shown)


jschwartzentruber (changes requested):
Functionally ... this is really slick. I had uMatrix block XHR to mozilla.org/allizom.org and it degraded nicely to text inputs. I appreciate the "loading" message during the request too.

I think the UI needs work though... currently, it looks like this (for "create"):
image
... or this (for "edit"):
image

In both cases, when the form is loaded it's unclear that you must choose a provider to get Product/Components populated. Could you auto-select the first provider? Loading one and having to change it is more obvious than having a blank field that unlocks the product/component selection.

I'd also rather have the "current" product and component be hidden fields, but with a "reset" button at the end of that line if the selection is changed from the original values.

The "reload" button is out-of-place under the "Product" heading. I think it should be included after the provider selection.

So here's my low-budget mock-up to summarize:
image

jschwartzentruber (approved):
After talking to @choller, I think my UX improvements would be easy enough for us to implement later. Your changes are already very functional and a big improvement over what we had.

PR #698, opened 2021-06-02
Allow browsing potential bug duplicates before filing a new external bug

### LINKED ISSUE #681: Bug suggestion when creating a new Bugzilla bug
The main goal is to provide a list of existing bugs matching a summary, when users are creating a bug through `crashes/<crashid>/createbug/`.

A new Vue.js component will be created that will use the user's Bugzilla Token from #676 and query the `possible_duplicates` endpoint.
If duplicates are found, a table listing them will be displayed below the summary, including these informations:
- Bug ID, with a link towards that bug
- Summary
- Component
- Status

An action button will be added on each line to assign the crash to that bug. This will require a new API endpoint to assign an existing Bugzilla Bug to a crash.

This endpoint would be available as `PUT rest/crashes/<crashid>/external-bug`, where the input serializer will expect:
- a bug provider ID
- an external bug ID
The endpoint will then simply create a new `Bug` instance for that provider and external bug id, and assigning the bug to the crash bucket (exactly like in `views.createExternalBug`)

If no existing bug is selected, the bug creation continues as before, using the `BugzillaProvider` class.

## References

- API doc for [possible_duplicates](https://bmo.readthedocs.io/en/latest/api/core/v1/bug.html?highlight=duplicates#possible-duplicates)
- Example [request](https://bugzilla-dev.allizom.org/rest/bug/possible_duplicates?product=Firefox&product=Core&product=Toolkit&summary=fenix+issue+with+sidebar&limit=12&include_fields=id%2Csummary%2Cstatus%2Cresolution%2Cupdate_token%2Ccc%2Ccomponent) - would need an authenticated account

## Current display on bugzilla

For reference, this is how it looks on Bugzilla when trying to fill a bug:

Screenshot_2021-05-25 Enter A Bug

EvaBardou:
Closes #681


server/crashmanager/serializers.py
```diff
@@ -199,3 +199,16 @@ class Meta:
 
     def get_mode(self, obj):
         return obj.mode.value
+
+
+class ExternalBugSerializer(serializers.Serializer):
+    provider = serializers.PrimaryKeyRelatedField(queryset=BugProvider.objects.all())
+    bug_id = serializers.CharField()
+    redirect_url = serializers.SerializerMethodField(required=False)
```
La0, line 207: I would use either `url` or `details_url` as this could be used in other cases

server/crashmanager/views.py
```diff
@@ -1034,6 +1034,35 @@ def partial_update(self, request, pk=None):
...
+        extBug, _ = Bug.objects.get_or_create(
+            externalId=serializer.validated_data.get('bug_id'),
+            defaults={
+                "externalType": serializer.validated_data.get('provider')
+            }
+        )
+        entry.bucket.bug = extBug
+        entry.bucket.save(update_fields=['bug'])
+
+        return Response(ExternalBugSerializer({'entry': entry, **serializer.validated_data}).data)
```
La0, line 1063: Please add a small comment to explain that you need to send the entry here (and that's why you need to use `update` instead of `perform_update`)

server/frontend/src/components/Bugs/PossibleDuplicates/List.vue
```diff
@@ -0,0 +1,160 @@
+<template>
+  <div>
+    <button
```
La0, line 3: Please add the input text to manage summary instead of the django form (using a bootstrap group input for display)

La0 (commented):
A few nits + the missing input, and we are good to go

La0:
Deployed on https://fuzzmanager.teklia.com/crashmanager/crashes/5940/createbug/

jschwartzentruber (commented):
Is there a reason we need a new API instead of implementing `partial_update` on `Bucket`? The bucket endpoint already returns `bug.externalId` as `bug`, so I'd rather add a field on the `Bucket` serializer to take a `bug_provider` (either `classname` or `hostname`) and do a PATCH of both `bug` and `bug_provider` instead... That matches the `CrashEntry` API more closely.

The UX looks good. I did get a 500 error when clicking `Use this bug` button.

jschwartzentruber:
To explain further ... on `CrashEntry` we have a number of related models that we don't expose on the REST API directly, but treat them only as implementation details (`Product`, `OS`, `Platform`, `TestCase`, etc.). We expose the data for these models inline on the crashes serializer for `list`/`view`, implicitly create missing models on `create`, and update the linked `crash.testcase.quality` on `partial_update` of `testcase_quality` (on the `CrashEntry`).

We already started down this path on the `Bucket` serializer too ... as the `.bug` field on the buckets endpoint returns not `bucket.bug.pk`, but `bucket.bug.externalId`. I think it would be more consistent to do the linking operation as a `partial_update` on `Bucket` which expects `bug_provider` to exist and implicitly creates a `Bug` object.

La0:
Sure, we can do that, using `BucketViewset.partial_update` and adding a `bug_provider` foreign key field on `BucketSerializer`.

It seemed easier to implement a dedicated endpoint, but you're right it looks cleaner from the API point of view.

EvaBardou:
> Is there a reason we need a new API instead of implementing `partial_update` on `Bucket`? The bucket endpoint already returns `bug.externalId` as `bug`, so I'd rather add a field on the `Bucket` serializer to take a `bug_provider` (either `classname` or `hostname` ... I guess using pk, since there is already a `BugProvider` serializer) and do a PATCH of both `bug` and `bug_provider` instead... That matches the `CrashEntry` API more closely.
> 
> The UX looks good. I did get a 500 error when clicking `Use this bug` button.

@jschwartzentruber I rebased the PR and updated it to use the `BucketViewSet` instead of creating a new one to assign the external Bugzilla bug, could you please review again? Thx!

P.S.: I also added a fix for another 500 error on signature list (mode: `all`) page. It occurred when the `BugProvider` linked to a `Bucket.bug` didn't have an `urlTemplate` (like on #697).

server/crashmanager/templates/signatures/index.html
```diff
@@ -37,8 +37,10 @@
             <td>{{ entry.size }}</td>
             <td>{{ entry.quality }}</td>
             <td>
-                {% if entry.bug %}
+                {% if entry.bug.externalType.urlTemplate %}
```
La0, line 40: Please keep the `{% if entry.bug %}` check too, or it might fail

server/crashmanager/serializers.py
```diff
@@ -124,15 +124,20 @@ def create(self, attrs):
 
 
 class BucketSerializer(serializers.ModelSerializer):
+    signature = serializers.CharField(style={'base_template': 'textarea.html'}, required=False)
     bug = serializers.CharField(source='bug.externalId', default=None)
     # write_only here means don't try to read it automatically in super().to_representation()
     # size and best_quality are annotations, so must be set manually
     size = serializers.IntegerField(write_only=True, required=False)
     best_quality = serializers.IntegerField(write_only=True, required=False)
+    provider = serializers.PrimaryKeyRelatedField(write_only=True, required=False, queryset=BugProvider.objects.all())
```
jschwartzentruber, line 133: nit: I'd prefer this field to be `bug_provider`

jschwartzentruber (approved):
Looks good besides naming nit

PR #701, opened 2021-06-04
Vueify Bugzilla bug publication

### LINKED ISSUE #695: Vueification of Bugzilla bug publication
We'll work on this before #694 

The main goals of this work are:
- remove HTML rendering related to comment publication from the Python `Provider` implementations
- use Vue.js component to publish Bugzilla comments instead of Python code
- interpolate templates using [Mustache](https://github.com/janl/mustache.js/) in the frontend, instead of the variable substitution from the backend.

Python methods removed:
- `BugzillaProvider.handlePOSTCreate`
- `BugzillaREST.createBug`

Bugzilla API calls needed in Javascript:
- [bug publication](https://bmo.readthedocs.io/en/latest/api/core/v1/bug.html?highlight=bug%20create#create-bug)
- [attachment publication](https://bmo.readthedocs.io/en/latest/api/core/v1/attachment.html?highlight=attachment#create-attachment) (should be done in #694 )

Data needed:
 - all the fields from `BugzillaTemplate` except `comment`:
   - summary
   -  whiteboard
   - keywords
   - op_sys
   - platform
   - cc
   - assigned_to
   - priority
   - severity
   - alias
   - qa_contact
   - status
   - resolution
   - version
   - target_milestone
   - attributes
   - description
   - security
   - security group
 - crash entry fields:
   - metadata
   - short signature
   - test case
   - crash data
   - args
   - OS
   - platform
   - client

Rendered fields using data provided by the form are:
- description
- comment
- attrs
They'll be rendered using [mustache.js](https://github.com/janl/mustache.js/)

Notes:
- test case is displayed and editable only when non-binary
- test case filename is editable
- rendered fields are pre-populated from the template (and updated when switching of templates), but can be edited

A dedicated API endpoint should be developed to serialiaze a crash entry and provided cleaned up data for the form (it would be read-only).

The `views.createExternalBug` method will support `GET` requests only, rendering a new HTML template using a new Vue.js component

The component would allow a user to:
- select a provider
- select a bugzilla bug template
- all fields are filled with data from the template
- retrieve automatically details about a crash entry
- display rendered fields when using a templatewith mustache.js (in dedicated parallel fields)
- input all fields from current form (mentionned above)

On submit, the component would query the Bugzilla API to:
- create a bug, and display errors if any,
- create attachments for crash data and test case (when present), and display errors if any
- finally display a link toward the bugzilla bug

A new API endpoint (`POST` on `rest/crashes/<id>/bug`) would be created to store metadata relevant to the comment, its serializer would use as input:
- crash entry ID
- provider ID
- bug ID on the provider (so the bugzilla bgu ID)
- rendered text comment
This endpoint would then call a new method on the Bug provider `hook_bug_created`, passing all received data from the frontend.
The Bugzilla implementation would do create a `Bug` instance for that provider and Bugzilla bug ID, and assign it to the crash bucket.

Here is a mockup of the new page:
mockup_bug

### Implementation steps:
- [x] 1. Create new overall Vue.js component with empty form fields and embed existing components to list products, components and find similar bugs.
- [x] 2. build configuration panel with provider & template selection
- [x] 3. fill fields on template selection
- [x] 4. live render of bug description using mustache
- [x] 5. publish bug on Bugzilla using REST api + available token
- [x] 6. call API on fuzzmanager backend to trigger hooks
- [x] 7. support test case (download file when needed) and publish as attachment.
- [x] 8. support crash data and publish as attachment.
- [x] 9. cleanup python provider code (and rest api)
- [ ] 10. (optional?) support `attrs` using clean key/values as set of inputs

Steps 1 to 6 could be done in a single Merge request,  then one merge request for each following steps.

EvaBardou:
Depends on #698 

References #695

Draft PR for now since I wasn't able to test my implementation, especially the bug publication by calling Bugzilla REST API (Bugzilla Dev down today).
**Edit (7th June):** Tested now, I successfully published my bug on Bugzilla Dev.


server/crashmanager/templates/bugzilla/create_external_bug.html
```diff
@@ -0,0 +1,11 @@
...
+
+{% block body_content %}
+<bugpublicationform
+    :provider-id="{{ provider }}"
+    provider-hostname="{{ hostname }}"
+    :template-id="{{ template }}"
+    :entry-id="{{ entry.id }}"
+    :bucket-id="{{ entry.bucket.id }}"
+/>
+{% endblock body_content %}
```
pyoor, line 11: minor nit - missing newline.

server/crashmanager/views.py
```diff
@@ -735,6 +735,40 @@ def createExternalBug(request, crashid):
...
+    if request.method == 'GET':
+        if 'provider' in request.GET:
+            provider = get_object_or_404(BugProvider, pk=request.GET['provider'])
+        else:
+            user = User.get_or_create_restricted(request.user)[0]
+            provider = get_object_or_404(BugProvider, pk=user.defaultProviderId)
+
+        template = provider.getInstance().getTemplateForUser(request, entry)
+
+        # attachmentData = {}
```
pyoor, line 755: Just curious - what's the reason for leaving these commented lines in place?

server/crashmanager/views.py
```diff
@@ -735,6 +735,40 @@ def createExternalBug(request, crashid):
...
+
+        # if template:
+        #     attachmentData = self.substituteTemplateForCrash(template, entry)
```
pyoor, line 758: Just curious - what's the reason for leaving these commented lines in place?

pyoor (approved):
lgtm

server/crashmanager/views.py
EvaBardou: I guess that was a way for me to remember that I will have to handle `attachmentData` coming from `subsituteTemplateForCrash` in the next PR (which will support the test case). That's cleaner to remove it though, will do in the next commit.

PR #702, opened 2021-06-09
Publish testcase and crashdata files as attachments during Bugzilla bug publication

EvaBardou:
Depends on #701 

References #695


pyoor (approved):
:+1:

PR #704, opened 2021-06-10
Vueify Bugzilla comment publication

### LINKED ISSUE #694: Vueification of Bugzilla comment publication
The main goals of this work are:
- remove HTML rendering related to comment publication from the Python `Provider` implementations
- use Vue.js component to publish Bugzilla comments instead of Python code

Python methods removed:
- `BugzillaProvider.handlePOSTComment`
- `BugzillaREST.createComment`

Bugzilla API calls needed in Javascript:
- [comment publication](https://bmo.readthedocs.io/en/latest/api/core/v1/comment.html?highlight=comment#create-comments)
- [attachment publication](https://bmo.readthedocs.io/en/latest/api/core/v1/attachment.html?highlight=attachment#create-attachment)

Data needed:
 - bugzilla bug id to comment on
    -  from associated bucket (provided as a prop)
    - or from user
 - text comment from user
 - is the comment private (from user)
 - text comment from potential template
 - ~~crash data from user~~ crash data is not present in the current form, but supported by provider
 - test case from user

The `views.createExternalBugComment` method will support `GET` requests only, rendering a new HTML template using a new Vue.js component

The component would allow a user to:
- select a provider
- select a bugzilla bug id (if not already provided)
- select a bugzilla comment template
- display text input for comment
- display rendered comment when using a template (in a dedicated HTML paragraph, read-only)
- input test case as text (in a textarea)
- input private field (as a checkbox)

On submit, the component would query the Bugzilla API to:
- create a comment, and display errors if any,
- create attachments for crash data and test case (when present), and display errors if any
- finally display a link toward the bugzilla comment

A new API endpoint (`POST` on `rest/crashes/<id>/comment`) would be created to store metadata relevant to the comment, its serializer would use as input:
- crash entry ID
- provider ID
- comment ID on the provider (so the bugzilla comment ID)
- rendered text comment
This endpoint would then call a new method on the Bug provider `hook_comment_created`, passing all received data from the frontend.
The Bugzilla implementation would do nothing.

Here is a mockup of the new page:
mockup_comment_publish

EvaBardou:
Depends on #703 

Closes #694


server/frontend/src/components/Comments/PublicationForm.vue
```diff
@@ -0,0 +1,492 @@
...
+            Cancel
+          </button>
+        </div>
+      </form>
+    </div>
+  </div>
+</template>
+
+<script>
+var Mustache = require("mustache");
```
pyoor: Just curious but why not import?

server/frontend/src/components/Comments/PublicationForm.vue
```diff
@@ -0,0 +1,492 @@
...
+      ? this.entry.rawCrashData
+      : this.entry.rawStderr;
+    if (this.bugId) this.externalBugId = this.bugId;
+
+    let data = await api.listBugProviders();
+    this.providers = data.results.filter(
+      (p) => p.classname === "BugzillaProvider"
+    );
+    this.provider = this.providers.find((p) => p.id === this.providerId);
+    this.selectedProvider = this.provider.id;
```
pyoor: Could the situation ever occur where this page is accessed yet no BugzillaProvider has been registered?

pyoor (approved):
:+1:

server/frontend/src/components/Comments/PublicationForm.vue
EvaBardou: I kept the Django view code as is, it prevents accessing the component if no provider exists yet since it will raise a 404 error before rendering:
```
if request.method == 'GET':
    if 'provider' in request.GET:
        provider = get_object_or_404(BugProvider, pk=request.GET['provider'])
    else:
        user = User.get_or_create_restricted(request.user)[0]
        provider = get_object_or_404(BugProvider, pk=user.defaultProviderId)
```

server/frontend/src/components/Comments/PublicationForm.vue
EvaBardou: I guess we could use import, I just stupidly c/c the Mustache doc: [https://github.com/janl/mustache.js/#usage](https://github.com/janl/mustache.js/#usage)

Do you want me to update that?

PR #708, opened 2021-06-15
Vueify signature list

### LINKED ISSUE #696: Vueification of signatures list
A new read-only API viewset would be created to list and retrieve details of available `Signature`

Filters to support to match `views.signatures`:
- `tool` as a tool's name (not FK)
- `bug__externalId`
- `shortDescription__contains`
- `signature__contains`
- `optimizedSignature__isnull`
- text query through `q` (on `shortDescription__contains` & `signature__contains`)
- `ids` with a comma-separated list of ID

By default:
1. show non-triaged crashes (when `bug=None`), except if `all=1` is specified
2. limit to `user.defaultToolFilter` if any is set on the current usze

Sorts to support:
- by ID
- by description
- by bucket size (from annotation)

A new component should be created to list signatures, and keeping the same render. It would of course use the endpoint mentionned above.

The `views.signatures` view would be used to simply display the component. All logic would go through the API endpoint and `get_queryset` method.

EvaBardou:
Closes #696


server/crashmanager/tests/test_signatures.py
```diff
@@ -88,111 +39,13 @@ def test_signatures_no_login(client, name, kwargs):
...
-    LOG.debug(siglist)
-    assert len(siglist) == 2  # 2 buckets
-    assert siglist[0] == bucket2
-    assert siglist[0].size == 1
-    assert siglist[1] == bucket1
-    assert siglist[1].size == 3
-    assert_contains(response, "bucket #1")
-    assert_contains(response, "bucket #2")
-    assert_contains(response, ALL_SIGS_FMT % 2)
+    assert_contains(response, 'signatureslist')
```
La0, line 48: Please add a unit test for the new serializer, simply calling the API with `?vue=1` should give a different output

server/frontend/src/components/Signatures/List.vue
```diff
@@ -0,0 +1,262 @@
...
+      }
+      this.fetch();
+    },
+    sortBy(sortKey) {
+      const keyChange = this.sortKey !== sortKey;
+      this.reverse = !keyChange ? !this.reverse : false;
+      this.sortKey = sortKey;
+      this.fetch();
+    },
+    parseHash() {
```
La0, line 194: Nit: this could be a helper/mixins shared with the crash list

La0 (commented):
LGTM ! I'll let @pyoor and/or @jschwartzentruber review and test it too

pyoor (approved):
LGTM.  Thanks!

PR #716, opened 2021-06-16
Add a Vue component suggesting Bugzilla users matching a search string

### LINKED ISSUE #705: Suggest Bugzilla users for CC and assigned-to fields on bug publication
A new Vue.js component should be created to suggest Bugzilla users for CC and assigned-to fields on the Bug publication form.
This component should appear only if a `BugzillaProvider` is selected AND a bugzilla token is available for that provider (otherwise, stick to the current text input fields).

The Bugzilla endpoint to use is `/rest/user/suggest?match=<input>` (it seems undocumented on the [official doc](https://bmo.readthedocs.io/en/latest/search.html?q=suggest&check_keywords=yes&area=default), but is currently used by Bugzilla frontend).
The endpoint needs authentication, and will provide a list of suggested users, sorted by pertinence.
For each user, we should display:
- the `gravatar` image
- the `real_name`
- the `email`

This would exactly match the display on Bugzilla:
Screenshot 2021-06-11 at 09-17-45 1630566 - Crash

A prop on the component must control if the component support multiple users or a single one:
- the CC field can have multiple users
- the assigned-to field has one user (at most)

Once selected, you can simply display (and store) the user's email address: this is the current behaviour on Bugzilla.

If there is an existing value in the template, it should be an email address: do not make any query towards Bugzilla to check or resolve that user (we trust the template).

EvaBardou:
Closes #705

### DIFF
--- a/server/frontend/src/bugzilla_api.js
+++ b/server/frontend/src/bugzilla_api.js
@@ -25,6 +25,14 @@ export const fetchPossibleDuplicates = async ({ hostname, params, headers }) =>
     )
   ).data;
 
+export const fetchSuggestedUsers = async ({ hostname, params, headers }) =>
+  (
+    await bugzillaAxios.get(`https://${hostname}/rest/user`, {
+      params,
+      headers,
+    })
+  ).data;
+
 export const createBug = async ({ hostname, headers, ...data }) =>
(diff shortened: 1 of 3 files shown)


pyoor (approved):
Looks good to me.

PR #718, opened 2021-06-16
Fix and prevent some Bug/Comment publication errors

EvaBardou:


server/frontend/src/components/Bugs/Comments/PublicationForm.vue
```diff
@@ -257,6 +258,7 @@ export default {
     crashData: "",
   }),
   async mounted() {
+    Base64.extendString();
```
pyoor, line 261: I'm worried about extending the String prototype as this might cause issues later that aren't inherently clear.  Is there a reason you did this as opposed to `Base64.encode(str)`?

server/frontend/src/components/Bugs/Comments/PublicationForm.vue
```diff
@@ -205,6 +205,7 @@
 
 <script>
 var Mustache = require("mustache");
+import { Base64 } from "js-base64"; // eslint-disable-line import/no-unresolved
```
pyoor, line 208: Any idea why eslint can't detect this in package.json?

pyoor (changes requested):

server/frontend/src/components/Bugs/Comments/PublicationForm.vue
EvaBardou: No particular reason, you're right, I'll change for `Base64.encode(str)`.

server/frontend/src/components/Bugs/Comments/PublicationForm.vue
La0: My bad: eslint was a bit mis-configured, the `import/resolver` was overriding the webpack path (mostly because I do that on another project). The issue was that eslint did not find anymore the correct paths towards node-modules, from sub-components.

It's now fixed in 7586f1ce8b7eb330909c9e70f7671f1b31a9f48c : you can check by editing the `js-base64` import to an invalid one (say `js-basewhatever`), and it will then fail.

pyoor (approved):
Looks great.  Thanks!

PR #720, opened 2021-06-18
Refactor user settings form to build notifications

### LINKED ISSUE #710: Refactor user settings form to build notifications
Refs #706 

The current `usersettings` view is using several `POST` actions and could be simplified by using a single Django form:
- select tools for default filters
- select default provider
- select default template

Then we would extend that form to allow a user to configure:
- its email address (required to send email notifications)
    - this field would only be editable if `settings.ALLOW_EMAIL_EDITION=True` (default to True)
- which notifications the user wants to receive
    - one checkbox (boolean field) per notification type (we only have `inaccessible_bug` & `bucket_hit`)

## Steps

1. refactor user settings to use a Django `forms.Form`
2. add email field to User model & user settings
3. add notifications boolean fields to User model & user settings

EvaBardou:
Closes #710 

Preview of the new form:
Capture d’écran du 2021-06-18 15-54-35

**Note:** With the current implementation of the `User` model, accessing `crashmanager/usersettings/` from a newly created FuzzManager instance will crash.   
Default values on `defaultTemplateId` and `defaultProviderId` are set even if no `BugProvider` nor `Template` exists. It could be a good thing to refactor those fields to a [`GenericForeignKey`](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#django.contrib.contenttypes.fields.GenericForeignKey) for `defautlTemplateId` and a basic `ForeignKey` towards `BugProvider` model for `defaultProviderId`.   
Other possibility would be to simply remove default values on both fields and allow `null`/`blank` values.


pyoor (approved):
LGTM.

PR #721, opened 2021-06-23
Add django-notifications package + Create notifications in commands called by cron jobs

### LINKED ISSUE #711: Add django-notifications
Refs #706 

We would use the open-source Django app [django-notifications](https://pypi.org/project/django-notifications-hq/) to add a new Database model `Notification`.
This app is "sold" as a look-alike for Github notifications:
> django-notifications is for building something like Github “Notifications”

The new [model](https://github.com/django-notifications/django-notifications/blob/master/notifications/base/models.py#L140) allows us to store a notification for one or more users, with:
- a message decoupled in action, verb, time + optional target object
- if the message was already `emailed`

We would need to implement a method to get all potential recipients for a given notification, following these rules:
- all users watching a bucket (through `BucketWatch` model), when the bucket has a new hit
    - executed in `triage_new_crashes`
- all users with related bucket crashes in their toolfilter, when a bug is inaccessible by the backend Bugzilla user 
    - executed in `bug_update_status`

As `django-notifications` allows us to send notifications through a Django signal, the implementation would look like:

```python
# When receiving a new crash in a bucket
notify.send(
    recipient=bucket.watchers.all(),
    actor=bucket,
    verb="bucket_hit",
    target=crash_entry,
    level="INFO",
    description="the bucket XXX received a new crash entry YYY"
)
```

Each user would get a `Notification` instance with the relevant informations, allowing him to acknowledge his own notification.

EvaBardou:
Depends on #720 

Closes #711

### DIFF
--- a/server/crashmanager/management/commands/bug_update_status.py
+++ b/server/crashmanager/management/commands/bug_update_status.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.core.management import BaseCommand, CommandError
+from notifications.signals import notify
 
 from crashmanager.management.common import mgmt_lock_required
 from crashmanager.models import Bug, BugProvider
@@ -58,3 +59,18 @@ class Command(BaseCommand):
                     elif bug.closed is None:
                         bug.closed = bugStatus[bugId]
                         bug.save()
+
+            for bugId in bugIds:
+                if bugId not in bugStatus:
+                    bugs = providerBugs.filter(externalId=bugId)
+                    for bug in bugs:
+                        notify.send(
+                            bug,
+                            recipient=bug.tools_filter_users,
+                            actor=bug,
+                            verb="inaccessible_bug",
+                            target=bug,
+                            level="info",
+                            description=f"The bug {bug.pk} pointing to the external bug {bug.externalId}"
+                                        f" on {provider.hostname} has become inaccessible"
+                        )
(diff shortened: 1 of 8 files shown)


server/crashmanager/management/commands/bug_update_status.py
```diff
@@ -58,3 +59,18 @@ def handle(self, *args, **options):
...
+                    bugs = providerBugs.filter(externalId=bugId)
+                    for bug in bugs:
+                        notify.send(
+                            bug,
+                            recipient=bug.tools_filter_users,
+                            actor=bug,
+                            verb="inaccessible_bug",
+                            target=bug,
+                            level="info",
+                            description=f"The bug {bug.pk} pointing to the external bug {bug.externalId}"
```
pyoor, line 74:
```suggestion
                            description=f"Bucket {bug.pk} associated with bug {bug.externalId}"
```

pyoor (approved):

server/crashmanager/management/commands/bug_update_status.py
EvaBardou: That's quite confusing to have the Bug internal pk after "Bucket" (`Bucket {bug.pk}`) no?

La0: I opened #727 to followup on that, this does not seem blocking as it's "only" wording and can be easily updated at a later stage.

PR #723, opened 2021-06-25
Add an endpoint to list unread notifications + Add a Vue component to display those notifications

### LINKED ISSUE #712: Build inbox for notifications
Refs #706 

The `django-notifications` app provides an `unread()` queryset: we would use that in a new List API view exposed as `/rest/inbox` to list all unread notifications for the connected user. 
The serializer might be a bit hard to setup, as we need to support different related objects but it's [possible using different sub-serializers](https://github.com/django-notifications/django-notifications#serializing-the-django-notifications-model).

A new Vue.js component would also be implemented and served as a simple Details view on `/inbox` to display all notifications, alongside a **Dismiss** button (see below for details).

Each notification would be processed by the Vue.js component, displaying custom information for each `verb`, and using the related `actor` and `target` objects to build links toward pages for these objects:
- crash entry link
- bucket detail link

EvaBardou:
Depends on #721 

Closes #712


pyoor (approved):

server/crashmanager/views.py
```diff
@@ -145,6 +147,10 @@ def settings(request):
     return render(request, 'settings.html')
 
 
+def inbox(request):
```
La0, line 150: Please use a `TemplateView` here to stick to Class-based-views for new views

PR #725, opened 2021-06-28
Add push Notification support and retrieve unread notifications count every 60s

### LINKED ISSUE #714: Push Notification support
Refs #706 

The `django-notifications` app provides a template tag `notifications_unread` to get the number of notifications users for the current user.

We would simply display an icon with the number of notification when there are any notifications. The display would be updated while browsing on the website.

Regarding [push notifications](https://developer.mozilla.org/en-US/docs/Web/API/Push_API), we can implement them as a followup task (this does not change anything on the base implementation above)

A Vue.js component would be used on every page load to fetch the current number of notifications and any new notifications since the last fetch.
This could be done using the API needed to list unread notifications on the inbox page (using an optional `?after=timestamp` parameter)
The endpoint would be polled every X minutes: no need for a websocket implementation here, as it seems a tad overkill.

EvaBardou:
Depends on #724

Closes #714

### DIFF
--- a/server/server/settings.py
+++ b/server/server/settings.py
@@ -96,6 +96,7 @@ TEMPLATES = [
             'context_processors': [
                 'django.contrib.auth.context_processors.auth',
                 'django.template.context_processors.debug',
+                'django.template.context_processors.request',
                 'django.template.context_processors.i18n',
                 'django.template.context_processors.media',
                 'django.template.context_processors.static',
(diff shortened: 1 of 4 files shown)


pyoor (approved):

server/server/settings.py
```diff
@@ -96,6 +96,7 @@ def resolver_context_processor(request):
             'context_processors': [
                 'django.contrib.auth.context_processors.auth',
                 'django.template.context_processors.debug',
+                'django.template.context_processors.request',
```
La0, line 99: Please add a comment explaining why this change is needed (required for specific view from django-notifications IIRC)

PR #726, opened 2021-06-29
Add a new management command to send notifications by email

### LINKED ISSUE #715: Send mail notifications
Refs #706 

The app `django-notifications` does **NOT** implement any mail functionality, but it has an `emailed` field on the `Notification` model.

We would implement a new Django management command named `notify` that would:
- iterate over all notifications with `emailed=False`
- check that the linked user has set its email
- check that the linked user is still watching this notification type
- build an email body using a Django template
- send the email through Django builtin system (using instance settings)
- update the `emailed` field on the notification to `True`

It would be the responsibility of the instance administrators to run that management command regularly as a cron (every X minutes).

EvaBardou:
Depends on #725 

Closes #715


pyoor (approved):

server/crashmanager/templates/notification_mail.html
```diff
@@ -0,0 +1,9 @@
+{% autoescape off %}
+Hello {{ user.user.username }},
+
+You received a new notification on FuzzManager.
+Notification: {{ notification.description }}
+
+-----------
+FuzzManager
+{% endautoescape %}
```
pyoor, line 9: Minor nit - missing newline.

PR #739, opened 2021-07-13
Cast externalId to int when comparing result from Bugzilla

### LINKED ISSUE #736: Accessible bugs generate an "inaccessible" notification.
When `bug_update_status` is run, notifications should be generated for "inaccessible" bugs, if they are private and the automation account can't access the bug. Instead I'm seeing public bugs generate a notification. I believe bugzilla is truncating the bugId list.

jschwartzentruber:
Fixes #736

### DIFF
--- a/server/crashmanager/management/commands/bug_update_status.py
+++ b/server/crashmanager/management/commands/bug_update_status.py
@@ -61,7 +61,7 @@ class Command(BaseCommand):
                         bug.save()
 
             for bugId in bugIds:
-                if bugId not in bugStatus:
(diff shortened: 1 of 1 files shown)


server/crashmanager/management/commands/bug_update_status.py
```diff
@@ -61,7 +61,7 @@ def handle(self, *args, **options):
                         bug.save()
 
             for bugId in bugIds:
-                if bugId not in bugStatus:
+                if int(bugId) not in bugStatus:
```
sylvestre, line 64: oh, python... :)

PR #750, opened 2021-07-16
Excluding users who have already receive the same notification in bug_update_status command

### LINKED ISSUE #738: `inaccessible_bug` notifications are generated multiple times
In [`bug_update_status.py`](https://github.com/MozillaSecurity/FuzzManager/blob/5e25efd0da6bb4fc7f91aecde295e26a2008d2ac/server/crashmanager/management/commands/bug_update_status.py#L67), there's no check for existing notifications. Every time the command is run, a duplicate notification will be generated.

EvaBardou:
Closes #738

### DIFF
--- a/server/crashmanager/management/commands/bug_update_status.py
+++ b/server/crashmanager/management/commands/bug_update_status.py
@@ -1,5 +1,7 @@
 from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
 from django.core.management import BaseCommand, CommandError
+from notifications.models import Notification
 from notifications.signals import notify
 
 from crashmanager.management.common import mgmt_lock_required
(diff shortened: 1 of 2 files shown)


jschwartzentruber (approved):
Can you add this to the test too? I think this should work:

```
diff --git a/server/crashmanager/tests/test_mgmt_bug_update_status.py b/server/crashmanager/tests/test_mgmt_bug_update_status.py
index 0c0b655..4cc5458 100644
--- a/server/crashmanager/tests/test_mgmt_bug_update_status.py
+++ b/server/crashmanager/tests/test_mgmt_bug_update_status.py
@@ -67,3 +67,8 @@ def test_fake_with_notification(mock_get_bug_status):
     assert notification.description == f"The bug {bug.pk} pointing to the external bug {bug.externalId}" \
                                        f" on {bug.externalType.hostname} has become inaccessible"
     assert notification.target == bug
+
+    # calling again should not generate a duplicate notification
+    call_command("bug_update_status")
+
+    assert Notification.objects.count() == 1
```

PR #756, opened 2021-07-19
Fix binary testcases loading/encoding on browser side

### LINKED ISSUE #755: Binary test cases are not encoded properly by the browser
This issue has been mentionned on #749 and on the chat.

Binary test cases are not properly base64-encoded when being sent to Bugzilla.

EvaBardou:
Closes #755

### DIFF
--- a/server/frontend/src/components/Bugs/PublicationForm.vue
+++ b/server/frontend/src/components/Bugs/PublicationForm.vue
@@ -780,12 +780,14 @@ export default {
           let content = this.testCaseContent;
           // If the testcase is binary we need to download it first
           if (this.entry.testcase_isbinary) {
-            content = await api.retrieveCrashTestCase(this.entry.id);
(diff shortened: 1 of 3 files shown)


server/frontend/src/components/Bugs/PublicationForm.vue
```diff
@@ -780,12 +780,15 @@ export default {
           let content = this.testCaseContent;
           // If the testcase is binary we need to download it first
           if (this.entry.testcase_isbinary) {
-            content = await api.retrieveCrashTestCase(this.entry.id);
+            const resp = await api.retrieveCrashTestCaseBinary(this.entry.id);
+            content = new Uint8Array(resp.data);
```
La0, line 784: You should directly do that step in `retrieveCrashTestCaseBinary` so the internal API stays consistent (we get usable payloads from the `api.js` module)

La0 (changes requested):

PR #762, opened 2021-07-23
Fix bucket watch notifications

jschwartzentruber:
Merge the two triage functions (celery/cron) into one, and generate watch notification in `CrashEntry.save` so it is generated regardless of how the crash is assigned to the bucket.


pyoor (approved):
:+1:

PR #763, opened 2021-07-26
[crashmanager] Use pointer cursor in crashmanager rows

pyoor:


jschwartzentruber (changes requested):
This has the same effect as removing the class entirely. Maybe `cursor:cell;`?

PR #775, opened 2021-08-06
Update old task runs.

jschwartzentruber:
If there are multiple runs for a task, only the most recent is relevant. We won't update these until the expected `max_run_time` is exceeded, which can be misleading in TaskManager.


pyoor (approved):
:+1:

PR #782, opened 2021-08-13
Add pagination on notifications inbox

### LINKED ISSUE #741: Notifications should be paginated.

EvaBardou:
Closes #741


jschwartzentruber (changes requested):
This is good, except that dismissing any or all notifications doesn't invalidate the counts displayed. 

For `dismissAll` you could just reset all the counts to 0.
For individual `dismiss`, I'm not sure how the sub-component invalidates the parent. Is this possible?

jschwartzentruber (approved):
Perfect. Thanks!

PR #783, opened 2021-08-19
Add help popovers for substitution fields when publishing Bugzilla Bug/Comment

### LINKED ISSUE #693: Add help for substitution fields available in bug templates
We support a number of substitutions in bug templates. These should be exposed as a guide in the template editor, as a help text somehow.

https://github.com/MozillaSecurity/FuzzManager/blob/495e9cd2a01f289cb6fd487da216cc8485198468/server/crashmanager/Bugtracker/BugzillaProvider.py#L173

EvaBardou:
Closes #693


pyoor (approved):
This looks good.  Thanks!

pyoor (approved):

jschwartzentruber (approved):
I actually like the original using bootstrap-vue, but should wait until after #786 is fixed. This is better in the meantime.

PR #784, opened 2021-08-24
Add tooltips for allowed params in CrashEntry/Bucket lists search query

### LINKED ISSUE #746: Add tooltip hint or similar to search query displaying allowed parameters
When using the advanced query to search crashes or buckets, it would be helpful to list the available parameters and their expected types.

EvaBardou:
Based on #783 

Closes #746

### DIFF
--- a/server/frontend/package.json
+++ b/server/frontend/package.json
@@ -10,7 +10,6 @@
     },
     "dependencies": {
         "axios": "^0.21.1",
-        "bootstrap": "^5.1.0",
         "handlebars": "^4.7.7",
         "js-base64": "^3.6.1",
         "lodash": "^4.17.21",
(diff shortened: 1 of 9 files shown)


server/frontend/package.json
```diff
@@ -10,6 +10,8 @@
     },
     "dependencies": {
         "axios": "^0.21.1",
+        "bootstrap": "^5.1.0",
```
La0, line 13: Beware this will embed bootstrap into the JS build too...

pyoor (commented):
It looks like you've also committed the changes from #783 here as well.

jschwartzentruber (approved):
This is good. Thanks!

PR #795, opened 2021-09-17
Show graph of recent bucket activity

jschwartzentruber:
This required augmenting the signatures rest API to return a count of crash history within that bucket (respecting tool filter).

It's impossible to test the performance of this without a complete dataset, so I'll watch/revert on deployment if the performance is bad. My local testing with 7k crashes across ~100 buckets showed no slowdown.

image

image


server/frontend/src/components/ActivityGraph.vue
```diff
@@ -0,0 +1,119 @@
...
+   * fm_data: Array[ Object { begin: String (ISO date), count: Number } ]
+   * start: Number (timestamp in ms)
+   * step: Number (ms)
+   * end: Number (timestamp in ms)
+   * clip: Number
+   *
+   * returns: Array[ Object { begin: Number (timestamp in ms), count: Number } ]
+   */
+  const result = [];
+  let cp = start,
```
pyoor, line 27: nit `const`

pyoor, line 27:
```suggestion
  const cp = start,
```

pyoor (approved):
lgtm!

server/frontend/src/components/ActivityGraph.vue
jschwartzentruber: `+=` would fail on line 35 if it were `const`.

PR #807, opened 2021-11-04
Combine wordbreak and responsive tables

pyoor:
Allows table-layout `auto` to decide when to apply horizontal scroll or word break.

image
image


jschwartzentruber (approved):
This is a big improvement. It doesn't full un-break taskmanager at low screen widths. I think we need to add `class=table-responsive` there in a follow-up.

jschwartzentruber:
deployed

PR #809, opened 2021-11-08
Crash list UI changes

jschwartzentruber:
There are three changes here:
- fix line-wrapping (no fields wrap except date, short-signature, and version).
- make testcase column more concise by replacing `(binary)` with an icon, and making size human-readable.
- truncate short-signature and version after two lines.

N.B. #808 also changes the date-time format, not shown here.

Before:
image

After:
image


pyoor (approved):
Look great!

choller (approved):

PR #810, opened 2021-11-08
Make testcase size it's own column.

jschwartzentruber:
Split out testcase size. This makes size a sortable value.

Before:
image

After:
image


pyoor (approved):
lgtm!

tysmith (approved):
+1

PR #813, opened 2021-11-10
Fix CovReporter --preprocessed argument handling.

jschwartzentruber:


pyoor (approved):
Thanks for fixing this.

PR #828, opened 2022-02-23
Repack wheel to exclude 'extra_requires'

jschwartzentruber:
fuzzing-decision (used by taskmanager) uses a direct url reference (allowed by [PEP 508](https://www.python.org/dev/peps/pep-0508/)). But PyPI cannot accept direct url references.

We don't need any of the extras anyways for the wheel (Collector/FTB only).

See https://github.com/pypa/twine/issues/430

### DIFF
--- a/repack_wheel.sh
+++ b/repack_wheel.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+set -e -x
+
+#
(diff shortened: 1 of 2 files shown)


jschwartzentruber:
I'm not sure if there's a better place to put the repack script....

tysmith (approved):
lgtm

PR #776, opened 2021-08-11
Groundwork for mypy support in FuzzManager

nth10sd:
This PR lays the groundwork for mypy support in FuzzManager.

- In the `misc/` folder, I had to rename `afl-libfuzzer` to `afl_libfuzzer` as mypy does not like dashes
- I took the mypy config in [Lithium](https://github.com/MozillaSecurity/lithium/) and adapted it here
- I added mypy and other `types-*` packages to the `requirements3.0.txt` file

GitHub Actions CI [is here](https://github.com/nth10sd/FuzzManager/actions).

Current master branch has 3188 mypy errors after installing all the *-stubs packages (and their latest versions on a Python 3.10 venv), this PR now has 1558 errors left. (51.13% out of 3188 fixed)

PR #829, opened 2022-03-14
Add mypy type information

nth10sd:
Continued here from #776 because the `black` code formatting rev bb12ebd389864bbba8c2c9e15ab808efae54e660 made merging/rebasing too painful, so I squashed the changes and rebased as best I can over in this PR.

458 mypy issues left, initially it was 3,101, 85% progress so far. Tested on mypy 0.940.


jschwartzentruber:
ICYMI https://github.com/nth10sd/FuzzManager/pull/1

nth10sd:
Argh, sorry I missed that.

PR #831, opened 2022-04-13
Improve memory usage for bucket edit reassignment

jschwartzentruber:
Instead of chunking the entire query, fetch all IDs and chunk those.

Otherwise `LIMIT`/`OFFSET` chunking can degrade SQL performance enough to cause timeout.


tysmith (approved):
LGTM

PR #863, opened 2022-12-08
Bump to Django 4.1.4

### LINKED ISSUE #857: Upgrade to Django 3.2
This is required to kickstart the GCP migration.

We need to upgrade the FuzzManager app to Django 3.2.

Steps:
1. check release notes for breaking changes from 2.2 to 3.2
2. identify potential issues
3. upgrade directly to 3.2
4. identify issues & django dependant plugins that may break

The first goal is to identify all issues and build a clear path for an upgrade. Then we'll be able to upgrade data & code.

EvaBardou:
Closes #857 

Closes #832


EvaBardou:
@jschwartzentruber The linting task was broken because the Flake8 `repo` was still pointing to the Gitlab reference so I updated it to point towards Github instead. 
It's still failing though, because of the `yeska` hook but I saw that you are aware of the issue (https://github.com/asottile/yesqa/issues/150).

Also, a new setting must be provided with Django 4, I didn't find any obvious logic on how you override Django settings while deploying (through env variables, TC secret or whatever) but I guess you might just be adding a new `settings_*.py` to override desired values on your instances as done in `server/server/settings_docker.py`? 
If so, please don't forget to add a new `CSRF_TRUSTED_ORIGINS` entry containing a list of trusted origins. For more information, see [Django official documentation](https://docs.djangoproject.com/en/4.0/ref/settings/#std-setting-CSRF_TRUSTED_ORIGINS).
If not, don't hesitate to tell me how I should refactor my changes to support custom assignations.

jschwartzentruber:
> @jschwartzentruber The linting task was broken because the Flake8 `repo` was still pointing to the Gitlab reference so I updated it to point towards Github instead.
> It's still failing though, because of the `yeska` hook but I saw that you are aware of the issue ([asottile/yesqa#150](https://github.com/asottile/yesqa/issues/150)).

This is caused by the flake8 section in tox.ini. Inline comments have always been non-standard, and flake8 recently stopped supporting them. I'll fix this in master.

> I guess you might just be adding a new `settings_*.py` to override desired values on your instances as done in `server/server/settings_docker.py`?

Yes, we have all settings in a settings.py file that lives only on the server.

jschwartzentruber (approved):
I'll squash and merge this once #864 lands.

jschwartzentruber:
> I think we do want to keep python 3.7 support for the client libraries only until it's unsupported upstream (in 6 months...). It's fine if the server requires 3.8 or later.

@EvaBardou btw, I have a solution for this. I'll open a PR to re-add 3.7 client tests after this lands.

jschwartzentruber:
Merged in #865

EvaBardou:
> @EvaBardou btw, I have a solution for this. I'll open a PR to re-add 3.7 client tests after this lands.

Oh perfect, thank you!

PR #868, opened 2022-12-14
Update CI & README

jschwartzentruber:
Add Python 3.11. Update git cloning (fetch by revision rather than branch, because it's more reliable). Update README for recent Django upgrades.


jschwartzentruber:
Python 3.11 tests are client only because of #869

pyoor (approved):

PR #870, opened 2022-12-22
Add task failure notifications

jschwartzentruber:
This notifies all users who have "view taskmanager" permission whenever a pool has a failed task (rate limited to 24h). LIke all notifications, it's opt-in. There's no way currently to make it more granular by user, so we'll get notifications for pools we don't care about too.


server/frontend/src/components/Notifications/TasksFailed.vue
```diff
@@ -0,0 +1,71 @@
...
+      </a>
+    </div>
+  </div>
+</template>
+
+<script>
+import { errorParser, formatClientTimestamp } from "../../helpers";
+import * as api from "../../api";
+
+export default {
```
pyoor, line 26: This is fine for now but we may want to create a generic class for all notifications and extend them where needed,

pyoor (approved):

Issue #939, opened 2023-11-21
Coverage Report page should show the date


mozfreddyb:
When going to an unpublished coverage Report Summary, e.g., covmanager/collections/xyz/summary/, it should show the date of said run.

If the stars align (and the django models have the right data somewhere nearby), I can try and give this a try, so I will assign it to myself.

jschwartzentruber:
Fixed in https://github.com/MozillaSecurity/FuzzManager/commit/505b12a83afe1c624c4cf7167e5351c52aeefcb6

PR #940, opened 2023-11-22
WIP: Add 'created' field to reportsummary and expose in UI (#939)

mozfreddyb:
This only adds a _created_ field during cache generation, which
wouldn't make it work for old, cached reports.
It also involved quite a bit of vue hackery, due to the data
not being available during django page render.
Maybe the model ought to change to take not only a `collectionid`
but also a `collection.created`?
This is also WIP because I'm literally dumping the Python `created`
attribute into the JSON and reading it out in JS. There is no explicit
serialization/parsing of the date and I'm not sure what format
is going to be used. Heh.

### DIFF
--- a/server/covmanager/tasks.py
+++ b/server/covmanager/tasks.py
@@ -125,6 +125,7 @@ def calculate_report_summary(pk):
         elif not data:
             # This is the root
             data = coverage
+            data["created"] = collection.created
         else:
             summary.cached_result = json.dumps(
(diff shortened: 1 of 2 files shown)


jschwartzentruber:
Fixed in https://github.com/MozillaSecurity/FuzzManager/commit/505b12a83afe1c624c4cf7167e5351c52aeefcb6

PR #952, opened 2024-01-09
Zoneinfo backports is only required for python < 3.9

pyoor:


pyoor:
Not necessary when using `pip install -c requirements.txt`.

PR #954, opened 2024-01-09
Escape short signature

jschwartzentruber:
Use escape sequences for control and special block unicode characters

### DIFF
--- a/FTB/Signatures/tests/test_CrashInfo.py
+++ b/FTB/Signatures/tests/test_CrashInfo.py
@@ -28,6 +28,7 @@ from FTB.Signatures.CrashInfo import (
     NoCrashInfo,
     RustCrashInfo,
     int32,
+    unicode_escape_result,
 )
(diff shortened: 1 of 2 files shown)


FTB/Signatures/tests/test_CrashInfo.py
```diff
@@ -3008,3 +3009,13 @@ def test_SanitizerHardRssLimitHeapProfile():
...
+
+
+def test_unicode_escape():
+    """test that unicode special and control characters are escaped"""
+
+    @unicode_escape_result
+    def testfunc():
+        return """\ufffd\u008dA"""
+
+    assert testfunc() == r"""\u{fffd}\u{8d}A"""
```
tysmith, line 3021: Can you add a unicode char that is not expected to be escaped as well?

tysmith (approved):

pyoor (approved):

PR #1033, opened 2024-12-05
Add default directories and settings for full UI testing

jschwartzentruber:


jschwartzentruber:
These cause errors in some corners of the UI if they don't exist.

tysmith (approved):

PR #1043, opened 2025-01-08
Move covmanager to frontend

jschwartzentruber:
Merge https://github.com/MozillaSecurity/FuzzManager-dev/pull/16


server/frontend/src/api.js
```diff
@@ -64,3 +64,43 @@ export const listPools = async (params) =>
 
 export const listTasks = async (params) =>
   (await mainAxios.get("/taskmanager/rest/tasks/", { params })).data;
+
+export const covManagerDiffStats = async ({ path, params, cb }) =>
+  await mainAxios
```
pyoor, line 69: You should only call await inside of an async function if you need the result.  Otherwise return the promise.

server/frontend/src/components/Covmanager/Patch.vue
```diff
@@ -0,0 +1,318 @@
...
+          if (response.ok) {
+            return response.json();
+          }
+          swal("Oops", E_SERVER_ERROR, "error");
+        })
+        .then((json) => {
+          this.collections = json["results"];
+          let analysis_results = [];
+
+          let self = this;
```
pyoor, line 246: Copying references to this can be avoided by using arrow functions.

pyoor (approved):
Looks good.  I left a few notes but none that need to be addressed in this PR.

PR #1060, opened 2025-03-07
Signatures view is sometimes very slow

gitstart:
This PR was created by [GitStart](https://gitstart.com/) to address the requirements from this ticket: [FUZ-21](https://clients.gitstart.com/mozilla/12016/tickets/FUZ-21).

 --- 

## Description

This PR significantly improves the performance of the `/crashmanager/signatures/` view by introducing a new BucketStatistics model that maintains pre-calculated statistics for each bucket+tool combination.

## **Key Changes**

- Added `BucketStatistics` model to track `size` and `quality` metrics
- Implemented Django signals to keep statistics up-to-date
- Modified `BucketAnnotateFilterBackend` to use pre-calculated statistics
- Added management command for statistics population/rebuild

## **Performance Impact**

- Eliminates expensive JOIN operations with CrashEntry during list view
- Moves computation cost from read-time to write-time
- Maintains existing security model and tool filtering
- Supports efficient multi-tool aggregation

**Existing Complexity**

- O(b \* c) where b = buckets and c = crash entries per bucket

**New Complexity**

- O(b \* t) where b = buckets and t = tools (we expect tools to be significantly smaller than c, hence the improvement)

## **Usage**

Current migration file, does not handle populating statistics for existing records. We created a management command for it:

`python manage.py populate_bucket_statistics`

`python manage.py populate_bucket_statistics --force` # Delete existing, and rebuild all statistics

### Testing

- Tested with Pattern A distribution (100 buckets with 10,000+ crashes each)
- Validated statistics accuracy after creating/deleting crash entry

### Demo

<https://www.loom.com/share/a745de5eb88742108ff83475b251c485?sid=de68e641-db4d-4ae7-b90c-5d7198252cc5>

### Sped Doc:

<https://docs.google.com/document/d/1EhbqCFu3EXCzEKp8bp8JvawRW5rMbY9Q7G2zOLu0iCI/edit?usp=sharing>


server/crashmanager/migrations/0018_bucketstatistics_and_more.py
```diff
@@ -0,0 +1,29 @@
...
+                ('last_update', models.DateTimeField(auto_now=True)),
+                ('bucket', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='crashmanager.bucket')),
+                ('tool', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='crashmanager.tool')),
+            ],
+        ),
+        migrations.AddConstraint(
+            model_name='bucketstatistics',
+            constraint=models.UniqueConstraint(fields=('bucket', 'tool'), name='unique_bucketstats_per_tool'),
+        ),
+    ]
```
jschwartzentruber, line 29: A Python operation should be added here to populate the bucket statistics, so that `python manage.py migrate` is enough to create the stats for existing data. I'm not sure if you can call management commands from here, the logic may need to be duplicated.

server/crashmanager/views.py
```diff
@@ -986,8 +1000,30 @@ class BucketAnnotateFilterBackend(BaseFilterBackend):
...
+
+        stats = stats.values('bucket').annotate(
+            total_size=Sum('size'),
+            min_quality=Min('quality')
+        )
+
         return queryset.annotate(
-            size=Count("crashentry"), quality=Min("crashentry__testcase__quality")
+            size=Coalesce(Subquery(stats.values('total_size')[:1]), 0),
+            quality=Subquery(stats.values('min_quality')[:1])
```
jschwartzentruber, line 1026: It would be nice if the subquery could be executed only once, but I don't think Django makes this possible.

jschwartzentruber (changes requested):
Please run `pre-commit run -a`. There are a number of formatting and other issues to fix.

server/crashmanager/models.py
```diff
@@ -390,6 +391,54 @@ def optimizeSignature(self, unbucketed_entries):
 
         return (optimizedSignature, matchingEntries)
 
+class BucketStatistics(models.Model):
+    bucket = models.ForeignKey(Bucket, on_delete=models.CASCADE)
+    tool = models.ForeignKey(Tool, on_delete=models.CASCADE)
+    size = models.IntegerField(default=0)
+    quality = models.IntegerField(null=True)
+    last_update = models.DateTimeField(auto_now=True)
```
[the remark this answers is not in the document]
gitstart-mozilla: This is noted.

server/crashmanager/models.py
```diff
@@ -390,6 +391,54 @@ def optimizeSignature(self, unbucketed_entries):
...
+                stats.quality = quality
+        stats.save()
+
+    @classmethod
+    def decrement_count(cls, bucket_id, tool_id):
+        stats = cls.objects.filter(
+            bucket_id=bucket_id,
+            tool_id=tool_id
+        ).first()
+        if stats and stats.size > 0:
```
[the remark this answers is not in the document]
gitstart-mozilla: Thanks for highlighting, will update.

server/crashmanager/migrations/0018_bucketstatistics_and_more.py
gitstart-mozilla: This is noted, we would explore calling the management command from here.

server/crashmanager/views.py
gitstart-mozilla: Hi @jschwartzentruber,
we explored this, and currently we haven't found an elegant provision by the django orm to combine subqueries.

PR #1062, opened 2025-03-10
FUZ-22 - API Token improvements - IP Restriction

gitstart:
This PR was created by [GitStart](https://gitstart.com/) to address the requirements from this ticket: [FUZ-22](https://clients.gitstart.com/mozilla/12016/tickets/FUZ-22).

 --- 

## Summary

This PR implements IP-based restrictions for API token authentication, enhancing security by allowing administrators to limit which IP addresses or ranges can use a given authentication token.

## Problem

Currently, API tokens provide access from any IP address, which poses a potential security risk if a token is compromised. There's no way to restrict token usage to specific network ranges.

## Solution

This PR introduces a new authentication mechanism that validates both the token and the client's IP address against a list of allowed IP ranges. Key components:

1. New `IPRestrictedTokenAuthentication` class that extends the standard `TokenAuthentication`
2. `TokenIPRestriction` model to store allowed IP ranges (CIDR notation) for each token
3. Management command to easily configure IP restrictions for tokens
4. Default "allow all" policy (0.0.0.0/0 and ::/0) for backward compatibility

## Implementation Details

- Each token can have multiple allowed IP ranges defined in CIDR notation

- Newly created tokens automatically get a default open policy

- The system supports both IPv4 and IPv6 restrictions

- Existing tokens are automatically migrated to have open access by default

## Usage

Administrators can manage token IP restrictions using the new management command:

```
python manage.py manage_token_restrictions <token_key> --add 192.168.1.0/24
python manage.py manage_token_restrictions <token_key> --remove 192.168.1.0/24
python manage.py manage_token_restrictions <token_key> --reset
python manage.py manage_token_restrictions <token_key> --list
```

## Testing

The PR includes tests that verify:

- Authentication succeeds for requests from allowed IP addresses

- Authentication fails for requests from restricted IP addresses

## Demo

<https://www.loom.com/share/f401143c73954279aa1956b248d1822f?sid=b054e5e9-bf75-4e8a-8e6e-311b5e495eaa>


server/server/management/commands/manage_token_restrictions.py
```diff
@@ -0,0 +1,128 @@
...
+        """Validate CIDR notation"""
+        try:
+            IPv4Network(cidr)
+            return True
+        except (AddressValueError, ValueError):
+            try:
+                IPv6Network(cidr)
+                return True
+            except (AddressValueError, ValueError):
+                return False
```
jschwartzentruber, line 32: Since this is also needed by the `TokenIPRestriction` model, it should be moved there as a `staticmethod`

server/crashmanager/views.py
```diff
@@ -1616,19 +1617,7 @@ class BugzillaTemplateEditView(UpdateView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-
```
jschwartzentruber, line 1619: All removals in this file look like a bad merge.

jschwartzentruber (changes requested):
Please run pre-commit and fix an errors.

server/crashmanager/views.py
gitstart-mozilla: This is noted @jschwartzentruber

server/server/management/commands/manage_token_restrictions.py
```diff
@@ -0,0 +1,154 @@
...
+                if confirm.lower() == "y":
+                    if open_ipv4:
+                        TokenIPRestriction.objects.filter(
+                            token=token, ip_range="0.0.0.0/0"
+                        ).delete()
+                    if open_ipv6:
+                        TokenIPRestriction.objects.filter(
+                            token=token, ip_range="::/0"
+                        ).delete()
+                    self.stdout.write("Open policy removed")
```
jschwartzentruber, line 124: If `--force` is given, this will not remove the default open restrictions.

jschwartzentruber (changes requested):
Just one minor problem remaining.

server/server/management/commands/manage_token_restrictions.py
gitstart-mozilla: Thanks for noting that, we would update to cover that scenerio.

PR #1071, opened 2025-03-21
fix: ensure crash triage happens when bucket changes

jschwartzentruber:
This eliminates the need to periodically check for this condition and run bulk updates which put load on the database.


tysmith (approved):

jschwartzentruber:
Regression from #1060

PR #1042, opened 2024-12-31
Redesign user settings UI

jschwartzentruber:
Merge https://github.com/MozillaSecurity/FuzzManager-dev/pull/15

Demo video: https://www.loom.com/share/5f0f300cad1b4924bb824fc762460bb0

PR #1079, opened 2025-04-01
Fix email and tool filter editing

### LINKED ISSUE #1078: User settings view clears the email field
Saving user settings results in empty `email` field in the Django `User` record. This makes it impossible for the user to login, resulting in 500 errors.

jschwartzentruber:
Fixes #1078


tysmith (approved):

jschwartzentruber:
Regression from #1042

PR #1085, opened 2025-04-08
Reporter cleanup

jschwartzentruber:
- use specific exception types (still inherit from RuntimeError for legacy code)
- add 429 to retry codes
- clean up arg types for TaskStatusReporter

### DIFF
--- a/TaskStatusReporter/TaskStatusReporter.py
+++ b/TaskStatusReporter/TaskStatusReporter.py
@@ -204,7 +204,8 @@ def main(argv=None):
                     )
                 time.sleep(opts.keep_reporting + random_offset)
         else:
-            report = opts.report_file.read_text()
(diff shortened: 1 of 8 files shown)


TaskStatusReporter/TaskStatusReporter.py
```diff
@@ -199,8 +204,7 @@ def main(argv=None):
                     )
                 time.sleep(opts.keep_reporting + random_offset)
         else:
-            with open(opts.report_file) as f:
-                report = f.read()
+            report = opts.report_file.read_text()
```
tysmith, line 207: We should likely lock here as well.

tysmith (changes requested):

tysmith (approved):

PR #1169, opened 2026-04-29
Enable more ruff

tysmith:
This set of rules will help maintain a high standard of code quality and act as guardrails for LLMs.

### DIFF
--- a/Collector/Collector.py
+++ b/Collector/Collector.py
@@ -772,7 +772,7 @@ def main(args: list[str] | None = None) -> int:
                 "Command line arguments:",
                 " ".join(args),
             )
-            print("")
+            print()
 
         if retJSON.get("env"):
             env = json.loads(retJSON["env"])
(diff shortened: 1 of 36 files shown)


Collector/Collector.py
```diff
@@ -83,7 +83,7 @@ def refreshFromZip(self, zipFileName: str) -> None:
 
             # Now clean the signature directory, only deleting signatures and metadata
             for sigFile in os.listdir(self.sigCacheDir):
-                if sigFile.endswith(".signature") or sigFile.endswith(".metadata"):
+                if sigFile.endswith((".signature", ".metadata")):
```
jschwartzentruber, line 86: TIL .. this should all be pathlib, but that's another problem

FTB/Running/GDB.py
```diff
@@ -87,7 +87,7 @@ def printImportantRegisters() -> None:
         ]
     elif isARM64():
         # ARM64 has GPRs from x0 to x30
-        regs = ["x" + str(x) for x in range(0, 31)]
+        regs = ["x" + str(x) for x in range(31)]
```
jschwartzentruber, line 90: I know this is all auto-fixes, but this should be an f-string (here and in RegisterHelper)

jschwartzentruber (approved):
lgtm besides suggestions

FTB/Running/GDB.py
tysmith: Yes there are a lot of potential f-string fixes. They can be addressed in another PR.

